PR #5, opened 2019-08-05
Cp from kubeflow

kkasravi:
move to v3 instead of v2

This repo is being kept insync with kubeflow/kubeflow/bootstrap/ (the kfctl cmd and pkg).
Each time we need to sync from kubeflow/kubeflow/bootstrap - we just need to call hack/cp_update. 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/5)


jlewi:
@kkasravi Can you provide a little more explanation for this PR?

What happens the next time we sync the changes from kubeflow/kubeflow to kubeflow/kfctl?

kkasravi:
@jlewi i've added the prow_config.yaml and ksonnet components {workflows, kfctl_go_test} until tektoncd is working

jlewi:
@kkasravi I think you need to update the tests if you want to actually test the code in kubeflow/kfctl. I think right now you are testing the code in kubeflow/kubefl

I've attached the logs from the build-train-deploy step

This is one log line that indicates the tests aren't working as desired

```
2019-08-07T21:59:35.219827360Z  main            rootdir: /mnt/test-data-volume/kubeflow-kfctl-presubmit-kfctl-go-5-20ddabf-7696-5c7b/src/kubeflow/kubeflow/testing/kfctl, inifile:
```

It looks like we are still checking out kubeflow/kubeflow and actually running the test from within that directory.

Should we commit this PR as is to get the V3 changes in and fix the tests in a follow on PR?

[kubeflow-kfctl-presubmit-kfctl-go-5-20ddabf-7696-5c7b-2334095401.log.txt](https://github.com/kubeflow/kfctl/files/3488189/kubeflow-kfctl-presubmit-kfctl-go-5-20ddabf-7696-5c7b-2334095401.log.txt)

jlewi:
/lgtm
/approve
/hold (so you see my comment; but just remove the hold).

I think we should stick with copying over kubeflow/kubeflow/testing to kubeflow/kfctl. I think we should defer any refactoring until we have switched over to kubeflow/kfctl

In terms of how to better componentize our tests and improve reusability; I don't know what the answer is. I would suggest deferring until we have a clearer issue of what the end state should be.

I suspect a part of the answer will be TektonCD and defining reusable Tekton tasks. 

Another reason not to invest in much ksonnet refactoring is that we want to get rid of ksonnet in our tests as soon as possible.

kkasravi:
@jlewi thanks

> In terms of how to better componentize our tests and improve reusability; I don't know what the answer is. I would suggest deferring until we have a clearer issue of what the end state should be.
> I suspect a part of the answer will be TektonCD and defining reusable Tekton tasks.

TektonCD is in progress, i've been experimenting with having TektonCD execute go code for e2e similar to the e2e go code that  kubernetes-sigs/application has (see https://github.com/kubernetes-sigs/application/blob/master/e2e/main_test.go#L57)

PR #10, opened 2019-08-16
[WIP] Tektoncd pipelines

kkasravi:
adding e2d tests using tektoncd pipelines

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/10)


jlewi:
I did a quick drive by; I was a little surprised by
https://raw.githubusercontent.com/kkasravi/manifests/e2e-pipelineruns/kfdef/gcp/kfctl-gcp-e2e-pipelinerun.yaml

Since that file is being referenced from prow_config as py_func to run a TektonPipeline; I was expecting a Tekton resource specifying the E2E test to run. Instead I see a KFDef.

It looks you are relying on KFDef to install an application which is then triggering the E2E workflow?

I'll take another look at your doc to see if its explained there.

kkasravi:
> It looks you are relying on KFDef to install an application which is then triggering the E2E workflow?

the e2e workflow includes the following manifests
- application-crds
- application
- tektoncd-crds
- tektoncd-install
- e2e-tasks
- e2e-pipelines
- e2e-pipelineruns

e2e-pipelines holds a kfctl config file and params that comprise the use case and tests.

jlewi:
@kkasravi How should we go about testing and migrating to Tekton? Right now E2E tests are written using ksonnet which returns an Argo workflow.

My first question is should we change from Argo and ksonnet to Tekton + pyfunc at the same time? Or should we replace ksonnet first and migrate to Tekton later?

I think it will take some time (i.e. post 1.0) before we have validated that Tekton is working. In addition to updating the test infra and run_e2e_workflow.py, we need to make sure we have good processes around debugging failures like we do for Argo.

So if we choose to migrate all at once that means we would need to continue to maintain and modify our existing ksonnet E2E tests workflows for some time.

Would it make sense to migrate our existing E2E test to a py_func first (maybe doing something like Jinja)?

Looking at 
https://github.com/kubeflow/kubeflow/blob/master/testing/workflows/components/kfctl_go_test.jsonnet

There's a fair bit of jsonnet/ksonnet use so rewriting as a PyFunc might take some time.

Some of the patterns thought might be portable to when we switch to Tekton. e.g. one of the things that we are doing is defining a template for a pod which then gets reused across all steps in the Argo workflow.

So we could do something like

* Define a string literal in Python containing a base PodTemplateSpec
* Define a helper function to combine that PodTemplateSpec with a specific command
   * i.e. Define a python function buildTemplate equivalent to the [jsonnet function](https://github.com/kubeflow/kubeflow/blob/5b652b18892c7b49af48ecc365b0af1080e0cb06/testing/workflows/components/kfctl_go_test.jsonnet#L242)

kkasravi:
> My first question is should we change from Argo and ksonnet to Tekton + pyfunc at the same time? Or should we replace ksonnet first and migrate to Tekton later?

I'm far enough down the tekton path that going to tekton + pyfunc is what I would recommend.
However, we should work out the details in the [Continuous Updating of Kubeflow Kustomize Manifests](https://docs.google.com/document/d/1oaBBJerOkKIuAAn_Swbu8uVBPtaYgvAVbQV1NxTsSSg/edit#). I've been doing what [update_jupyter_web_app.py](https://github.com/kubeflow/kubeflow/blob/master/py/kubeflow/kubeflow/ci/update_jupyter_web_app.py) does for centraldashboard but with tekton - as you suggested. The doc goes into a bit of detail within the TODO sections where you asked a number of questions. We may want to discuss in a short design meeting. I can demo the functionality. Currently I'm taking parts of update_jupyter_web_app.py and integrating them into the tekton pipeline. The tekton PipelineRun and Pipeline are treated as apps in kfdef, so the jupyter ci script ends up being a pipeline defined in a kfctl config file.

jlewi:
Why use KfDef and applications to define Tekton pipelines? Isn't that just an extra layer of indirection?

Why not just have a spec defining a Pipeline?

My expectation is that someone could do 

```
kubectl apply -f  jupyter_cd.yaml
```

Where jupyter_cd.yaml is a PipelineRun (or Pipeline?) that would continuously build and update the Jupyter web app.

What is the next bit of functionality/code that will be ready for review?

kkasravi:
@jlewi a few reasons for going this route

- known way of development for app developers for both ci and e2e
  - tests are added by just adding an overlay to manifests
- composable
  - each task is an overlay, a new overlay is just added to the config file
- reusable
  - each overlay is used by multiple kf apps
- strengthens our own product
  - we still have open issues on 
    - how apps are added with dependencies, health of an app, application CRs
- well tested parameter model
  - familiar to app developers
- independent of prow or any other ci/cd system. 
  - our entry point to prow is py_func which runs kfctl.
- locality, the developer just runs 
  - kfctl --config <ci-app.yaml> or
  - kfctl --config <e2e-app.yaml>
- multi-cluster
  - the developer can run ci or e2e framework on minikube | kind, run pipeline on different cluster

kkasravi:
> What is the next bit of functionality/code that will be ready for review?

parity with update_jupyter_web_app.py. Remaining work
- switch to tekton ssh repos
- add edit, create PR logic from update_jupyter_web_app.py as steps in update-manifest-task
- add other kf components (currently centraldashboard, profile-controller) as kf config files
- change params and secrets from intel gcp to kubeflow-ci
- update_jupyter_web_app.py assumes a PR in kubeflow components is open, need to add tekton PR at start of pipeline and propagate it to downstream tasks

I could add relevant sections above to kubeflow/testing#450 and kubeflow/kubeflow#3781

jlewi:
Why do we need a new familiar way to create Kubernetes resources?

If our E2E tests are workflows then shouldnt the workflow/pipeline just be a first class concept? Why hide it by wrapping it in KfDEf?

Can we start with something simple? Could we create a simple Tekton task or pipeline just for building the docker image for the central dashboard? Could we just create a simple YAML file to define those resources and then check it in?

Then we can extend that with steps to update the manifests and finally create the PR?

Why not just have people work with Tekton Tasks/pipelines directly?

Those are Kubernetes resources so develop ers should already know how to create those?

If developers want to generate pipelines using overlays (and I'm not sure we will) can't we just use kustomize? Why do we need a new tool (kfctl) to generate?

kkasravi:
> Can we start with something simple? Could we create a simple Tekton task or pipeline just for building the docker image for the central dashboard? Could we just create a simple YAML file to define those resources and then check it in?

sure, so check in the yaml under kubeflow/py/kubeflow/kubeflow/ci and add the location as an arg to py_func in prow_config?

jlewi:
@kkasravi I think the first step would be a YAML file to build the centraldashboard. So I would place it here
https://github.com/kubeflow/kubeflow/tree/master/components/centraldashboard

For the first PR I don't even want integration into CI (e.g. py_func) and we can't trigger it from CI until we 

1. Install TektonCD on the test cluster
1. Add tekton support to run_e2e_workflow.py

The point of this is to create a concrete example of how we would start to define Tasks,TaskRef and/or Pipelines to begin to construct pipelines to auto-update our kustomize manifests.

kkasravi:
@jlewi see kubeflow/kubeflow#4112

jinchihe:
@kkasravi Seems we need to let the PR https://github.com/kubeflow/testing/pull/434 merged firstly?

kkasravi:
Hi @jinchihe 

> @kkasravi Seems we need to let the PR [kubeflow/testing#434](https://github.com/kubeflow/testing/pull/434) merged firstly?

most of these changes are inherited from the branch sync-with-kubeflow, which I'll open a PR as soon as  kubeflow/kubeflow#4197 is merged into kubeflow. This branch really just needs `jq` added to the kfctl Dockerfile - it's called in a pipeline task.

jlewi:
@kkasravi What are the next steps with this PR? Have the relevant bits already been submitted in separate PRs?

kkasravi:
@jlewi yes, i think what remains is adding tekton pipelines as an alternative to argo workflows for our e2e tests. I think this mostly involves kubeflow/testing and then adding e2e specific tests per repo (unless we're putting all e2e tests for all repos under kubeflow/testing

Closing this PR

PR #28, opened 2019-09-05
Build kfctl in Dockerfile with Alpine based image

NathanHowell:
I'm not sure if this is used, but the `Dockerfile` produces an image that is not functional. The build image is Debian based (glibc) but the final runtime image is Alpine based (musl). There are three possible fixes:
* Switch the build image to Alpine
* Add libc6-compat to the runtime image
* Switch the runtime image to Debian or Distroless

I've arbitrarily chosen the first option and fixed up the `sh` and `sleep` commands. Additionally, `ca-certificates` is needed for `kfctl init` to complete successfully.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/28)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,11 +2,11 @@
 # Builder
 # 
 # Create a go runtime suitable for building and testing kfctl
-ARG GOLANG_VERSION=1.12.7
+ARG ALPINE_VERSION=3.10.1
+ARG GOLANG_VERSION=1.12.9-alpine3.10
 FROM golang:$GOLANG_VERSION as builder
 
-RUN apt-get update
-RUN apt-get install -y git unzip
+RUN apk add --no-cache bash build-base git python unzip wget
 
 # junit report is used to conver go test output to junit for reporting
 RUN go get -u github.com/jstemmer/go-junit-report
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -48,7 +48,8 @@ RUN make build-kfctl
 # Final image base
 #
 
-FROM alpine:3.10.1 as barebones_base
+FROM alpine:$ALPINE_VERSION as barebones_base
+RUN apk add --no-cache ca-certificates  # needed by kfctl
```
kkasravi: how did you determine ca-certificates was needed by kfctl?

NathanHowell: @kkasravi  I'll update the comment, if an https uri is referenced in the Kubeflow app.yaml then Go's http client needs ca-certificates to validate the remote cert. In this case I was testing with the 1.6.2 sample config:

```
apiVersion: kfdef.apps.kubeflow.org/v1alpha1
kind: KfDef
metadata:
  name: kubeflow_app
  namespace: kubeflow
spec:
  repos:
    - name: manifests
      uri: https://github.com/kubeflow/manifests/archive/56e2fb15db286198f7a53723cb1fbfecf3fe83fb.tar.gz
    - name: kubeflow
      uri: https://github.com/kubeflow/kubeflow/archive/0dbd2550372c003ba69069aeee283bd59fb1341f.tar.gz
```

gabrielwen:
@kkasravi sorry I don't have context to compare between base images, could I rely on you for the review?  thanks.

kkasravi:
@NathanHowell I opened a PR on your branch to make some small changes
https://github.com/NathanHowell/kfctl/pull/1

kkasravi:
you'll also need to rebase from master
`git fetch upstream master`
`git rebase -i upstream/master`

PR #29, opened 2019-09-06
v1beta1 KfDef

gabrielwen:
Bootstrapping KfDef for v1beta1.

Relates to kubeflow/kubeflow#4002

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/29)


jlewi:
Thanks
LGTM

@yanniszark @kkasravi comments?

gabrielwen:
please help and take a look at condition types and reasons, I try to fit the new status into our new semantics.

kunmingg:
Will it be compatible with https://github.com/kubeflow/kubeflow/blob/v0.6-branch/bootstrap/config/kfctl_gcp_iap.0.6.2.yaml ?

gabrielwen:
no, this is the new version.  we will make kfctl to be backward compatible with v1alpha1 KfDef, tho.

kkasravi:
there is a merged [doc](https://github.com/rthallisey/hyperconverged-cluster-operator/blob/e6cee852f17fd3794ec0e747b1658c5d264ac6b2/docs/conditions.md) on conditions and ApplicationStatusCondition

gabrielwen:
@kkasravi I've updated the conditions part according to the link, could you take a second pass?

kkasravi:
thanks @gabrielwen 
/lgtm

gabrielwen:
@yanniszark will wait for your review.  please approve this when you think it looks good.

pkg/apis/apps/kfdef/v1beta1/application_types.go
```diff
@@ -0,0 +1,170 @@
...
+// Plugin can be used to customize the generation and deployment of Kubeflow
+type Plugin struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec *runtime.Object `json:"spec,omitempty"`
+}
+
+// Secret provides information about secrets needed to configure Kubeflow.
+// Secrets can be provided via references e.g. a URI so that they won't
```
yanniszark: In the SecretSource struct, I don't see a way to provide a secret with a URI.
However, I do think it's a valid use case.
Maybe we could add a repoSource which will be a RepoRef struct and find a secret from one of the repos? 

cc @jlewi

pkg/apis/apps/kfdef/v1beta1/application_types.go
```diff
@@ -0,0 +1,170 @@
...
+
+type RepoRef struct {
+	Name string `json:"name,omitempty"`
+	Path string `json:"path,omitempty"`
+}
+
+type NameValue struct {
+	Name         string `json:"name,omitempty"`
+	Value        string `json:"value,omitempty"`
+	InitRequired bool   `json:"initRequired,omitempty"`
```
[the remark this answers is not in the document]
jlewi: @yanniszark is correct; we should be able to get rid of this.

pkg/apis/apps/kfdef/v1beta1/application_types.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+// Plugin can be used to customize the generation and deployment of Kubeflow
+type Plugin struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec *runtime.Object `json:"spec,omitempty"`
```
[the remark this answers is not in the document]
jlewi: What would be the problem with including TypeMeta and ObjectMeta fields?

IIRC I had originally tried using RawExtension and that didn't work; I think because it doesn't define DeepCopy or possibly some other methods.

pkg/apis/apps/kfdef/v1beta1/application_types.go
jlewi: We can always add fields later. How about waiting to add a URI field until we have a concrete use case?

Secrets generally shouldn't be stored in repositories so why would we need the ability to reference a secret in a repository?

pkg/apis/apps/kfdef/v1beta1/application_types.go
```diff
@@ -0,0 +1,170 @@
...
+// Secrets can be provided via references e.g. a URI so that they won't
+// be serialized as part of the KfDefSpec which is intended to be written into source control.
+type Secret struct {
+	Name         string        `json:"name,omitempty"`
+	SecretSource *SecretSource `json:"secretSource,omitempty"`
+}
+
+type SecretSource struct {
+	LiteralSource *LiteralSource `json:"literalSource,omitempty"`
+	HashedSource  *HashedSource  `json:"hashedSource,omitempty"`
```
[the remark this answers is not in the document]
jlewi: @kunmingg Could we drop HashedSource? I think we are using this for BasicAuth for the password but can't we send the password using a LiteralSource and rely on https to encrypt it?

pkg/apis/apps/kfdef/v1beta1/application_types.go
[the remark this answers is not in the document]
kkasravi: see https://pursuit.purescript.org/packages/purescript-kubernetes/0.1.1/docs/Kubernetes.Api.Runtime

I believe this means we make it rawExtension but convert to runtime.Object when unmarshalling

pkg/apis/apps/kfdef/v1beta1/application_types.go
```diff
@@ -0,0 +1,170 @@
...
+	// Can use any URI understood by go-getter:
+	// https://github.com/hashicorp/go-getter/blob/master/README.md#installation-and-usage
+	Uri string `json:"uri,omitempty"`
+}
+
+// KfDefStatus defines the observed state of KfDef
+type KfDefStatus struct {
+	Conditions []KfDefCondition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,6,rep,name=conditions"`
+	// ReposCache is used to cache information about local caching of the URIs.
+	ReposCache map[string]RepoCache `json:"reposCache,omitempty"`
```
jlewi: @gabrielwen @yanniszark Should we make ReposCache a list instead of a Map and move name into RepoCache?

This has two advantages

1. RepoCache would be more consistent with Repo
1. RepoCache would contain all the relevant information about a cached repo. 
    * Right now the name is stored as the key in the Map. So you can't just pass around a RepoCache struct; you'd have to pass around the tuple (name, RepoCache).

pkg/apis/apps/kfdef/v1beta1/application_types.go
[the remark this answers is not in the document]
kkasravi: we should upgrade Value to be either a string or array as shown [here](https://github.com/tektoncd/pipeline/blob/master/pkg/apis/pipeline/v1alpha1/param_types.go#L89)

pkg/apis/apps/kfdef/v1beta1/application_types.go
```diff
@@ -0,0 +1,170 @@
...
+	// KfPluginsProgressing means kfctl is applying plugins.
+	KfPluginsProgressing KfDefConditionType = "PluginsProgressing"
+
+	// KfKustomizeProgressing means kfctl is running package manager.
+	KfKustomizeProgressing KfDefConditionType = "KustomizeProgressing"
+)
+
+type KfDefConditionReason string
+
+const (
```
nrchakradhar: Should these definitions include "KfDefConditionReason" similar to above for KfDefConditionType?

pkg/apis/apps/kfdef/v1beta1/application_types.go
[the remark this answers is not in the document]
gabrielwen: updated.

pkg/apis/apps/kfdef/v1beta1/application_types.go
gabrielwen: good catch, thanks!

pkg/apis/apps/kfdef/v1beta1/application_types.go
```diff
@@ -0,0 +1,170 @@
...
+	// InvalidSecretsReason indicates secret(s) were not valid.
+	InvalidSecretsReason = "InvalidSecrets"
+
+	// ApplyPluginsFailedReason indicates plugin(s) were not applied successfully.
+	ApplyPluginsFailedReason = "ApplyPluginsFailed"
+)
+
+type KfDefCondition struct {
+	// Type of deployment condition.
+	Type KfDefConditionType `json:"type" protobuf:"bytes,1,opt,name=type,casttype=KfDefConditionType"`
```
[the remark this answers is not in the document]
gabrielwen: removed.

pkg/apis/apps/kfdef/v1beta1/application_types.go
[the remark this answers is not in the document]
gabrielwen: would there be a risk of using runtime.Object?  It already includes `TypeMeta` for us to identify the object, right?

pkg/apis/apps/kfdef/v1beta1/application_types.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+type KfDefConditionType string
+
+const (
+	// KfAvailable means Kubeflow is serving.
+	KfAvailable KfDefConditionType = "Available"
+
+	// KfDegraded means functionality of Kubeflow is limited.
+	KfDegraded KfDefConditionType = "Degraded"
```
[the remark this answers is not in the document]
gabrielwen: From the [doc](https://github.com/rthallisey/hyperconverged-cluster-operator/blob/e6cee852f17fd3794ec0e747b1658c5d264ac6b2/docs/conditions.md):
- `KfAvailable = true && KfDegraded = true` means KF is serving (central dashboard is accessible) but the services are limited.  For example, we can use notebook but not pipelines.
- `KfAvailable = false && KfDegraded = true` means KF is not serving and part/all of services are not serving as well.
- `KfAvailable = true && KfDegraded = false` means KF is serving and all services are healthy.

pkg/apis/apps/kfdef/v1beta1/application_types.go
gabrielwen: I'm fine with using a list.  updated the definition.

pkg/apis/apps/kfdef/v1beta1/application_types.go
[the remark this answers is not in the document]
yanniszark: @jlewi using the Object interface means that we force the Plugins to implement the Object interface.

> IIRC I had originally tried using RawExtension and that didn't work; I think because it doesn't define DeepCopy or possibly some other methods.

Do you have more info on that.
My understanding is that RawExtension should be used for storing arbitrary JSON.
Relevant link: https://github.com/kubernetes-sigs/service-catalog/issues/911#issuecomment-305506879

@gabrielwen using runtime.Object requires plugins to implement the Object interface.

What we want to store there is arbitrary JSON and my understanding from my search is that RawExtension is what should be used.

pkg/apis/apps/kfdef/v1beta1/application_types.go
yanniszark: In that case, the comment should be updated to reflect that.

pkg/apis/apps/kfdef/v1beta1/application_types.go
yanniszark: @jlewi this is more in line with the Kubernetes API Conventions.
I'm also in favor of this.

pkg/apis/apps/kfdef/v1beta1/application_types.go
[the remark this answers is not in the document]
yanniszark: I believe this doc was mentioned by @kkasravi.
I don't have a problem with using the specific conditions, just wondering if they fit Kubeflow as well.
For example, I don't think we can detect a condition like "KF is serving but not all services are serving" in kfctl.
However, individual conditions do make sense.

In addition, the doc refers to a library that doesn't exist anymore (https://github.com/openshift-kni/operator-status).
I will update the design doc to refer to this doc.
@jlewi @kkasravi thoughts?

pkg/apis/apps/kfdef/v1beta1/application_types.go
```diff
@@ -0,0 +1,211 @@
...
+	case ParamTypeArray:
+		return json.Marshal(arrayOrString.ArrayVal)
+	default:
+		return []byte{}, fmt.Errorf("impossible ArrayOrString.Type: %q", arrayOrString.Type)
+	}
+}
+
+type NameValue struct {
+	Name  string        `json:"name,omitempty"`
+	Value ArrayOrString `json:"value,omitempty"`
```
yanniszark: Why did this field become array or string?
Do we have a use case for this?

gabrielwen: @kkasravi

pkg/apis/apps/kfdef/v1beta1/application_types.go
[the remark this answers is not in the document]
kunmingg: Should be ok to drop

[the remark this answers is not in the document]
gabrielwen: done.

pkg/apis/apps/kfdef/v1beta1/application_types.go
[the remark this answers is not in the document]
jlewi: What's the problem with expecting plugins to implement the Object interface? Don't we need that so that we do things like generate appropriate DeepCopy methods?

We are using type Object in the v1alpha1 spec and it is working as expected. Here's the sample GCPPluginSpec

```
  plugins:
  - name: gcp
    spec:
      auth:
        iap:
          oAuthClientId: 147474701642-rn0ksbu4eikorfk8j0uvknubbke57ri1.apps.googleusercontent.com
          oAuthClientSecret:
            name: CLIENT_SECRET
      createPipelinePersistentStorage: true
```

So whether its a runtime.Object or runtime.Extension doesn't appear to influence the user facing API; seems like . So I suggest keeping it as is and if you think it should be changed open.

You can look at the existing implementation;
https://github.com/kubeflow/kubeflow/tree/master/bootstrap/pkg/kfapp/gcp 

So it doesn't look like we had to implement any special methods.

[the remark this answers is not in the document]
jlewi: I think to @kkasravi 's by using runtime.Object we avoid having to write our own unmarshaling logic.

pkg/apis/apps/kfdef/v1beta1/application_types.go
gabrielwen: removed URI parts.

pkg/apis/apps/kfdef/v1beta1/application_types.go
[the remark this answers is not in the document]
jlewi: I think we should follow the doc only if the conditions make sense in our use case. 

I might also suggest holding off defining conditions until we are sure we need them; i.e. we are writing code to implement. I might suggest commenting out and explaining why or opening up an issue to define track what conditions we think we need.

We might want to hold off defining the conditions until we actually need them; i.e. are writing the code to implement them.

Adding conditions later doesn't break compatibility but removing ones could.

pkg/apis/apps/kfdef/v1beta1/application_types.go
```diff
@@ -0,0 +1,170 @@
...
+
+const (
+	// KfAvailable means Kubeflow is serving.
+	KfAvailable KfDefConditionType = "Available"
+
+	// KfDegraded means functionality of Kubeflow is limited.
+	KfDegraded KfDefConditionType = "Degraded"
+
+	// KfPluginsProgressing means kfctl is applying plugins.
+	KfPluginsProgressing KfDefConditionType = "PluginsProgressing"
```
jlewi: Do we need PluginsProcessing? Should we hold off defining it until we know we need it?

pkg/apis/apps/kfdef/v1beta1/application_types.go
```diff
@@ -0,0 +1,170 @@
...
+
+	// KfKustomizeProgressing means kfctl is running package manager.
+	KfKustomizeProgressing KfDefConditionType = "KustomizeProgressing"
```
jlewi: Instead of KustomizeProcessing should it be InstallingApplications?

Kustomize is a bit of an implementation detail.

pkg/apis/apps/kfdef/v1beta1/application_types.go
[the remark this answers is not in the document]
gabrielwen: ok, I removed most of conditions/reasons and only keep one or two placeholders that I think we definitely need them.

pkg/apis/apps/kfdef/v1beta1/application_types.go
gabrielwen: removed.

pkg/apis/apps/kfdef/v1beta1/application_types.go
gabrielwen: also removed them.

jlewi:
@yanniszark @kkasravi Can you please take another look?

@gabrielwen What are the remaining issues to be resolved? It looks like we are waiting on @kkasravi to answer the question about why value is StringOrArray?

@yanniszark @kkasravi Can you please prioritize reviewing this? Its critical we get the v1beta1 API finalized so we can get the implementation done in time for 0.7.

swiftdiaries:
Great discussion !
I've nothing to add to the comments.
Just a small question, is there a doc or issue tracking the steps to be done make kfctl backwards compatible?

gabrielwen:
> Great discussion !
> I've nothing to add to the comments.
> Just a small question, is there a doc or issue tracking the steps to be done make kfctl backwards compatible?

kubeflow/kubeflow#4002 is the tracking issue but it doesn't really have the steps to make it compatible.  After this PR we need another one to unify both paths (v1alpha1/v1beta1)

gabrielwen:
> @yanniszark @kkasravi Can you please take another look?
> 
> @gabrielwen What are the remaining issues to be resolved? It looks like we are waiting on @kkasravi to answer the question about why value is StringOrArray?
> 
> @yanniszark @kkasravi Can you please prioritize reviewing this? Its critical we get the v1beta1 API finalized so we can get the implementation done in time for 0.7.

yes, I think that is the only question.

jlewi:
@swiftdiaries I think to make kfctl backwards compatible we just need to write a conversion function to convert v1alpha1 specs to v1beta1 specs; and then we can just run that when kfctl runs.

pkg/apis/apps/kfdef/v1beta1/application_types.go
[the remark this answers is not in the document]
yanniszark: @jlewi an Object is supposed to contain TypeMeta information (https://godoc.org/k8s.io/apimachinery/pkg/runtime).

The Spec is not an Object, but part of an Object.

In other words, the Plugin struct is an Object but not the Spec field.
That's why I say it doesn't make sense to use Object for just the Spec.

This is why I propose we define Spec as RawExtension, which is also for arbitrary JSON.
Does that make sense?

pkg/apis/apps/kfdef/v1beta1/application_types.go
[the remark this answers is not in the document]
kkasravi: @yanniszark @gabrielwen if KfDef becomes a CRD, should we define a structured schema 
see 
- [Future of CRDs: Structural Schemas](https://kubernetes.io/blog/2019/06/20/crd-structural-schema/)
- [Vanilla CRD OpenAPI subset: structural schemas](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190425-structural-openapi.md)

kkasravi:
> @kkasravi to answer the question about why value is StringOrArray?

there are use cases where an array would be valuable.
If we allow lists to be passed to kf apps then params.env is not the best way to pass a list.
This approach (string or array-of-string) is used by tektoncd in their parameter passing implementation. We should try and have well typed information for app params

pkg/apis/apps/kfdef/v1beta1/application_types.go
kkasravi: tektoncd has a similar construct

within the config file you may want to pass data that represents a dictionary
eg
      parameters:
      - name: cpu-devices
        value:
        - skylake
        - broadwell
        - crest
      - name: tpu-devices
        - nvidia-tesla-t4
        - nvidia-tesla-p100
        - nvidia-tesla-t4-vws

yanniszark:
> Would it make sense to change Plugin to include data where this info can be
embedded?

@kkasravi the Spec of a plugin is already free-form data.
The author of a Plugin can freely choose the correct structure that fits their plugin, including the structure you mentioned.
Why would we force a certain structure upon all plugins beforehand? 

> there are use cases where an array would be valuable.
> If we allow lists to be passed to kf apps then params.env is not the best way to pass a list.

Has such a use-case come up though? Is there a Kubeflow application that needs a list?
I'm hesitant about this because if we start conservatively then AFAIK we can make it StringOrArray without breaking backwards compatibility.
However, if we make it StringOrArray from the start, then we won't be able to take it back without breaking compatibility.

The last outstanding issue is Object vs RawExtension (https://github.com/kubeflow/kfctl/pull/29#discussion_r323288805).

pkg/apis/apps/kfdef/v1beta1/application_types.go
jlewi: ArrayOrString seems unnecessarily complicated

```
type ArrayOrString struct {
	Type      ParamType // Represents the stored type of ArrayOrString.
	StringVal string
	ArrayVal  []string
}
```

If we want to support arrays why not just do

```
type NameValue struct {
     Name string
     Value[] string
}
```

A single string argument could be handled by the case where array is size one.

But given how we are using parameters does it even make sense to support arrays? We are using kustomize parameters and creating key-value pairs. Doesn't that only handle string scalars? What would it mean for kustomize to substitute an Array into the manifests.

My vote would be to optimize for simplicity. I think our existing use cases are covered by scalar strings

```
type NameValue struct {
     Name string
     Value string
}
```

If we don't have a use case for Arrays then I suggest we stick with that. If we need to support more types in the future we can do that e.g.

```
type NameValue struct {
     Name string
     Value string
     ArrayValue []string
     MapValue map[string]string
}
```

jlewi: From @kkasravi 
> If we allow lists to be passed to kf apps then params.env is not the best way to pass a list.
This approach (string or array-of-string) is used by tektoncd in their parameter passing implementation. We should try and have well typed information for app params

How would a list let help users specify environment variables?

pkg/apis/apps/kfdef/v1beta1/application_types.go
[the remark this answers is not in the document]
jlewi: @yanniszark  Those seem like implementation concerns. How does that affect the user facing API? As the example above shows for GCP the spec would be 

```
plugins:
  - name: gcp
    spec:
      auth:
        iap:
          oAuthClientId: 147474701642-rn0ksbu4eikorfk8j0uvknubbke57ri1.apps.googleusercontent.com
          oAuthClientSecret:
            name: CLIENT_SECRET
      createPipelinePersistentStorage: true
```

From a user perspective if Object and RawExtension mean "unstructured json" then what's the difference?

If we change it to RawExtension what changes about the Object?

If the choice of RawExtension vs. Object doesn't affect the user facing API then IMO its an implementation detail. In which case we should just stick with Object for now because that's what the existing code is using so we know it works. If there's a reason that using Extension is better we can change it later. If the YAML a user writes isn't impacted then that's not a backwards incompatible change.

The docs don't make this very clear
https://godoc.org/k8s.io/apimachinery/pkg/runtime#RawExtension

> RawExtension is used to hold extensions in external versions.

> To use this, make a field which has RawExtension as its type in your external, versioned struct, and Object in your internal struct. You also need to register your various plugin types.

These docs lead me to conclude that its not really impacting the user facing API; i.e. the Json/YAML writes but rather how serialization/deserialization works.

jlewi:
Responding to @kkasravi's question about how users would modify the DM templates. I think this is working as intended. If there are specific parameters we want to expose (e.g. workloadIdentity, pipelinepersistentstorage) the plugin owners can define an appropriate high level API.

I don't think we need to let users embed arbitrary json/YAML.

The underlying config files are DM config files. If users want to easily modify those and customize them in weird ways they can do that by copying that to an appropriate directory, modifying them, and running gcloud to create the GCP infra. They can  then just use kfctl to install Kubeflow on their existing cluster.

kkasravi:
but could we do more in this area? My point is we already create an opinionated cluster:
2 machines with 8 vCPUs and 30 GB of memory, tpu's disabled, 0 gpus.
 
Just as we're providing device overlays that can be configured so that jobs run on accelerated hardware, can we provide hooks that vendors could utilize to meet customer requirements?

gabrielwen:
/hold
place hold until runtime.Object/runtime.RawExtension is resolved.

pkg/apis/apps/kfdef/v1beta1/application_types.go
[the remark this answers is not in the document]
jlewi: @gabrielwen @yanniszark It looks like I might have been confused I thought we were using Object but looks like we are currently using RawExtension.
https://github.com/kubeflow/kubeflow/blob/9455ab572f4fe287db984860b7e9452e6ac60a87/bootstrap/pkg/apis/apps/kfdef/v1alpha1/application_types.go#L107

It looks like the implementation has changed since v1alpha1.


Either way I think we want to pick the implementation that results in AP thatI looks like the following

```
kind: kfDefGcpPlugin
apiVersion: v1beta1
metadata:
  name: gcp-plugin
  namespace: kubeflow
spec:
  auth:
    iap:
      oAuthClientId: 147474701642-rn0ksbu4eikorfk8j0uvknubbke57ri1.apps.googleusercontent.com
      oAuthClientSecret:
        name: CLIENT_SECRET
  createPipelinePersistentStorage: true
```
I think the way to resolve this is to write a unittest to verify the YAML ends up looking the way we want it.

We already have such a unit test
https://github.com/kubeflow/kubeflow/blob/9455ab572f4fe287db984860b7e9452e6ac60a87/bootstrap/pkg/apis/apps/kfdef/v1alpha1/application_types_test.go#L251

So my suggestion is to port over the unittest to v1beta1 and update the YAML test file to look like what we want. Once the unittest passes we are good to go.

pkg/apis/apps/kfdef/v1beta1/testdata/kfctl_plugin_test.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kfdef.apps.kubeflow.org/v1beta1
+kind: KfDef
+spec:
+  plugins:
+    - name: fakeplugin
```
jlewi: @gabrielwen Are you still working on this test? Can we make this look like the spec in your comment?

```
apiVersion: kfdef.apps.kubeflow.org/v1beta1
kind: KfDef
spec:
  plugins:
    - kind: kfDefPlugin
      apiVersion: v1beta1
      metadata:
        name: gcp-plugin
      spec:
        kind: kfGcpPlugin
        apiVersion: v1alpha1
        metadata:
          name: gcp-plugin-spec
          namespace: kubeflow
        auth:
          iap:
            oAuthClientId: 147474701642-rn0ksbu4eikorfk8j0uvknubbke57ri1.apps.googleusercontent.com
            oAuthClientSecret:
              name: CLIENT_SECRET
        createPipelinePersistentStorage: true
```

gabrielwen: yes, I am.  of course I can change that.  before that... @kkasravi could you give me some pointers on how to generate deepcopy codes?  I keep getting errors when running `make deepcopy`

jlewi: How does the existing code work? I don't believe for GCPPluginSpec in v1alpha1 we are currently generating DeepCopy code.

gabrielwen: I need deepcopy code for `KfDef` in v1beta1, though.

jlewi: @gabrielwen why?

gabrielwen: because you need runtime.Object implementation to register the types:
https://github.com/kubeflow/kfctl/blob/48f8e149d7ddde344b64c6bc5b4508a2adbc5ae3/pkg/apis/apps/kfdef/v1alpha1/register.go#L50

anyway, I found the file.  it's in $GOPATH/src

jlewi: Great.

gabrielwen: test updated.

pkg/apis/apps/kfdef/v1beta1/testdata/kfctl_plugin_test.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: kfdef.apps.kubeflow.org/v1beta1
+kind: KfDef
+spec:
+  plugins:
```
jlewi: LGTM.

gabrielwen:
@yanniszark @kkasravi could I have lgtm from both of you?

jlewi:
/lgtm
/approve
/hold
To give @yanniszark time to review

pkg/apis/apps/kfdef/v1beta1/application_types.go
```diff
@@ -0,0 +1,222 @@
...
+	Overlays   []string    `json:"overlays,omitempty"`
+	Parameters []NameValue `json:"parameters,omitempty"`
+}
+
+type RepoRef struct {
+	Name string `json:"name,omitempty"`
+	Path string `json:"path,omitempty"`
+}
+
+// ParamType indicates the type of an input parameter;
```
yanniszark: I think this is left from the StringOrArray commit.
Can we remove the ParamType type?

pkg/apis/apps/kfdef/v1beta1/application_types.go
```diff
@@ -0,0 +1,222 @@
...
+
+// Repo provides information about a repository providing config (e.g. kustomize packages,
+// Deployment manager configs, etc...)
+type Repo struct {
+	// Name is a name to identify the repository.
+	Name string `json:"name,omitempty"`
+	// URI where repository can be obtained.
+	// Can use any URI understood by go-getter:
+	// https://github.com/hashicorp/go-getter/blob/master/README.md#installation-and-usage
+	Uri string `json:"uri,omitempty"`
```
yanniszark: nit: Can we change this to `URI` since its an acronym?

pkg/apis/apps/kfdef/v1beta1/application_types.go
```diff
@@ -0,0 +1,222 @@
...
+		if err != nil {
+			log.Errorf("Could not unmarshal plugin %v to the provided type; error %v", pluginName, err)
+		}
+		return nil
+	}
+
+	return NewPluginNotFound(pluginName)
+}
+
+type PluginNotFound struct {
```
yanniszark: Should we make these into KFErrors?

pkg/apis/apps/kfdef/v1beta1/application_types_test.go
```diff
@@ -0,0 +1,99 @@
...
+		if !reflect.DeepEqual(actual, c.Expected) {
+			pGot, _ := Pformat(actual)
+			pWant, _ := Pformat(c.Expected)
+			t.Errorf("Error parsing plugin spec got;\n%v\nwant;\n%v", pGot, pWant)
+		}
+	}
+}
+
+// Pformat returns a pretty format output of any value.
+func Pformat(value interface{}) (string, error) {
```
yanniszark: I believe there is a Pretty Print function already that @jlewi commited.
https://github.com/kubeflow/kfctl/blob/master/pkg/utils/logging.go#L10

gabrielwen:
@yanniszark done, could you take a final look? :)

yanniszark:
@gabrielwen thanks for all the great work.
/lgtm

One last concern I had, which is worth mentioning now:

AFAIK, we don't have a good story for providing secrets for securing applications.
eg we probably want to configure MySQL, OIDC clients, Minio etc with good random passwords and not default ones like we do now.

Are we confident that we'll be able to have a good story for this without needing a new version?
/cc @jlewi

@gabrielwen @jlewi that's all from my side, please remove the hold when you believe everything is ready.

jlewi:
@yanniszark We can extend SecretSource if need be to allow people alternative methods of referencing secrets.

Issue #7, opened 2019-08-14
Make kubeflow/kfctl source of truth for kfctl development; turn down kfctl in kubeflow/kubeflow


jlewi:
Opening this issue to track migrating development of kfctl from kubeflow/kubeflow to kubeflow/kfctl

@kkasravi What are the next steps here? Are we waiting on E2E tests to be migrated to kubeflow/kfctl?

/assign @kkasravi

kkasravi:
Yes, I need to add the tests and fix the error you noted where we are actually running e2e on Kubeflow/kubeflow

jlewi:
@kkasravi Any update on this? When do you think we should migrate development over from kubeflow/kubeflow to kubeflow/kfctl?

It would be great to do this well in advance of trying to cut 0.7.

jlewi:
@kkasravi any update?

kkasravi:
@jlewi 


There is are several pending PRs that we should merge
- #16 sync with kubeflow/bootstrap
- #17 prow fix to build kfctl and run in the right repo
- #11 add kfctl_k8s_istio config in presubmit tests

#17 I'm working on - should be ready today sometime
#11 is dependent on #17
#16 can be merged

jlewi:
Fantastic. Once we start switching over to kubeflow/kfctl; I think we'll need to update the kubeflow/manifests tests?

kkasravi:
@jlewi yep - we'll need to 
- remove kfctl tests in kubeflow/kubeflow that are redundant. 
- update manifests/tests/workflows/

jlewi:
@kkasravi It looks like #16 and #17 have been merged. Does this mean we are ready to start moving development to kubeflow/kfctl?

kkasravi:
Yes. 

I’ve also moved config over to manifests.
We should probably switch over to that in the same timeframe. I’ll ping you on that PR later today.

jlewi:
@kkasravi Do you want to send a note to kubeflow-discuss so we can get a heads up to all folks working on kfctl who might have changes in flight? Do we also want to try to identify any PRs in flight that will be impacted?

kkasravi:
done - https://groups.google.com/forum/#!topic/kubeflow-discuss/2T-iiDCb5Q0

jlewi:
Thanks @kkasravi so can you provide some detail here about the plan

1. When are we going to switch to make kubeflow/kfctl the source of truth?
[...]

PR #30, opened 2019-09-06
Use custom image registry

xaniasd:
resolves kubeflow/kubeflow#3210

Introduces imageRegistry parameter in KfDef spec. If set, all generated kustomizations will use repositories from the specified registry. 

**NB** istio and application images are not affected

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/30)


jlewi:
@xaniasd thanks for taking this on. 

A couple things until #7 is resolved; development of kfctl should happen in kubeflow/kubeflow and not in kubeflow/kfctl.

Why introduce a new parameter ImageRegistry into KfDefSpec rather than introduce a new go binary  or subcommands to provide bulk manipulation of the docker images in the kustomize files?

If we introduce KfDefSpec.ImageRegistry and just rewrite the images is that going to give us the power we need?

For example, if users want to pull from a private repo or replicate their images to their own mirror then we need to find and output all the existing images so that they can be mirror'd.

xaniasd:
hi @jlewi, thanks for your feedback.

> A couple things until #7 is resolved; development of kfctl should happen in kubeflow/kubeflow and not in kubeflow/kfctl.

thanks, noted

> Why introduce a new parameter ImageRegistry into KfDefSpec rather than introduce a new go binary or subcommands to provide bulk manipulation of the docker images in the kustomize files?

my implementation aims for replacing images, as described in the issue. Also along the lines of what the [script](https://github.com/kubeflow/kubeflow/blob/master/scripts/gke/use_gcr_for_all_images.sh) does.

> If we introduce KfDefSpec.ImageRegistry and just rewrite the images is that going to give us the power we need?
> 
> For example, if users want to pull from a private repo or replicate their images to their own mirror then we need to find and output all the existing images so that they can be mirror'd.

I probably miss some background here so I can't really answer your questions. I did find the design doc of kfctl in kubeflow/manifests#339, but a bit late I guess, please excuse my ignorance. What are the bulk manipulations we have in mind? There aren't any specified in the design doc. Furthermore, this functionality is relevant to kustomize, which according to the doc is not considered a plugin, correct?

let me know how you want to proceed with this!

jlewi:
@xaniasd The functionality we'd like is something like the following

```
kfctl set-image-registry <old prefix> <new prefix>
```

So the behavior would be something like the following

```
cd ${KFAPP}
kfctl build ....
```
* This will create all the kustomize manifests in ${KFAPP}/kustomize

The user would then do something like

```
cd ${KFAPP}/kustomize
kfctl set-image-registry gcr.io/kubeflow-images-public  myprivate/registry
```

* This would then recourse through all the kustomization.YAML files and replace the old prefix with the new prefix.

Eventually we might provide built in support for this as part of KFDef. I don't think we want to tightly integrate into KFDef right now because we are still experimenting with the right semantics for image replacement.

For KFDef on the other hand we are trying to get the API to v1beta1 in 0.7 and then v1 in Q4. So we don't want to add new/experimental fields to it right now.

jlewi:
@xaniasd Think you might be able to take another stab at this this week?

xaniasd:
@jlewi I don't think I'll have time for it for the next 5-10 days, pretty busy agenda unfortunately.

jlewi:
Thanks @xaniasd 

@gyliu513 @jinchihe Any interest in picking this up since you will need it for POWER?

gyliu513:
@alongzhi can you help take over this?

FYI @jinchihe @jlewi

hemantha-kumara:
> * This would then recourse through all the kustomization.YAML
@jlewi All kustomization.yaml doesn't have images.name & images.newName 
for example 
- https://github.com/kubeflow/manifests/blob/master/metadata/base/kustomization.yaml
- https://github.com/kubeflow/manifests/blob/master/istio/istio/base/kustomization.yaml  
with existing approach how 
This mean, there are two approaches to solve this problem
1. Either submit all kustmization files with have images field for all docker images
2. Parse all deployment and statefulset .yaml to find name of the image and do replace using kustomize edit set image <imagenamename>=<newName>

alongzhi:
@gyliu513 @jlewi Okay, I'm looking at this one.

nrchakradhar:
@alongzhi @gyliu513 @jlewi
https://github.com/kubeflow/manifests/pull/363/files
This might help.

jlewi:
Thanks @nrchakradhar 
@xaniasd Are you planning on updating this PR per my comments above?

xaniasd:
@jlewi sure, I'll give it a try

xaniasd:
@jlewi @yanniszark 

here's another version in line with what we discussed, let me know what you think.

As soon as we're ok with this approach, I'll close this PR and submit another one on kubeflow/kubeflow only with the good stuff.

cmd/kfctl/cmd/set-image-name.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v2"
+)
+
+var filter string
+var flatten bool
+
+func init() {
+	setImageNameCmd.Flags().StringVarP(&filter, "filter", "f", "", "Only set name for images with matching prefix")
+	setImageNameCmd.Flags().BoolVar(&flatten, "flatten", false, "Set to true for registries not supporting hierarchical paths with more than two components")
```
jlewi: Why does this mean? Can you give me example command line syntax?

jlewi:
@xaniasd Thanks. Could you provide some examples of what the command line syntax is?

xaniasd:
hi @jlewi, sure thing. See below:

```bash
# replace registry domain if exists, append if not
# so mysql:latest - test.io/mysql:latest
# gcr.io/kubeflow-images-public/ingress-setup:latest -> test.io/kubeflow-images-public/ingress-setup:latest 
kfctl set-image-name test.io
# same as above, but only for image names starting with gcr.io
kfctl set-image-name test.io --filter gcr.io
# gcr.io/kubeflow-images-public/ingress-setup:latest -> test.io/my-repo/kubeflow-images-public/ingress-setup:latest
kfctl set-image-name test.io/my-repo 
# gcr.io/kubeflow-images-public/ingress-setup:latest -> test.io/my-repo/ingress-setup:latest
kfctl set-image-name --flatten test.io/my-repo
```

about the effect of --flatten, you can find some more examples in the unit tests

yanniszark:
@jlewi after discussing with @xaniasd I realized that not all image registries support subdirectories.
In fact, many support only a flat structure (eg Docker registry).
Because of that, I believe it's important to support flattening the images.

The algorithm more or less is:
1. User gives a registry (eg `test.io/kf-v062/` under which to move images.
2. For every image (eg `gcr.io/kubeflow-images/vizier/katib/katib-ui`) matching `filter`, strip the origin and:
    1. `flatten=true`: Keep only the last part of the path. (`test.io/kf-v062/katib-ui`)
    2. `flatten=false`: Keep the whole path. (`test.io/kf-v062/kubeflow-images/vizier/katib/katib-uikatib-ui`)

jlewi:
Thanks @xaniasd @yanniszark 
semantics LGTM.

yanniszark:
Great!
@xaniasd shall we take this to kubeflow/kubeflow?

jlewi:
Closing this PR because the code was merged into kubeflow/kubeflow. The code has now been migrated to kubeflow/kfctl and is here
https://github.com/kubeflow/kfctl/blob/master/cmd/kfctl/cmd/set-image-name.go

PR #31, opened 2019-09-10
[WIP] upgrade to kustomize3

kkasravi:


jlewi:
Should we be making this change in kubeflow/kfctl or kubeflow/kubeflow?

Per #7 it doesn't look like development has switched over from kubeflow/kubeflow yet.

PR #51, opened 2019-10-12
sync with kubeflow v0.6.0-rc.0-207-gad0893ee

kkasravi:
snapshot taken of kubeflow master at Sat Oct 19 11:56:41 PDT 2019

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/51)

PR #61, opened 2019-10-24
Copy ci package from kubeflow to kfctl

richardsliu:
Attempting to fix https://github.com/kubeflow/kfctl/pull/51#issuecomment-545594365

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/61)

### DIFF
--- a/ci/application_util.py
+++ b/ci/application_util.py
@@ -1,0 +1,97 @@
+"""Utilities for updating various Kubeflow applications."""
+
+import logging
+import os
+import pathlib
+import tempfile
+
+from kubeflow.testing import util # pylint: disable=no-name-in-module
+
+import yaml
+
+def set_kustomize_image(kustomize_file, image_name, image):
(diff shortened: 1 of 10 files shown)


kkasravi:
the go.mod file has a dependency on 
```
        github.com/kubeflow/kubeflow/components/profile-controller/v2 v2.0.0
```
which I don't think we need anymore.

Here's how I would 'sync' with the latest from kubeflow/kubeflow as well as get the ci changes in

1. Make sure kfctl and kubeflow repos are siblings in your local repos - which is also the case in testing
2. cd to the kfctl repo
3. run hack/cp_update.sh - this copies over the changes from kubeflow and changes things like the imports
4. add the kubeflow/py/kubeflow/kubeflow/ci stuff
5. do `make build-ktctl` and `make build-kfctl-container` to see if it builds on mac and in docker

PR #76, opened 2019-11-06
update coordinator/converters/gcp API for kfctlserver

kunmingg:
This PR is Part of https://github.com/kubeflow/kubeflow/pull/4399: only include changes on kfctl side.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/76)


kunmingg:
@jlewi 
Will update unit tests once got lgtm on API changes

gabrielwen:
since you are removing `UseIstio` field, do you mind also remove it here?
https://github.com/kubeflow/kfctl/blob/08a2e3736b2ca25f1f7a327090ecc48e014aa90b/pkg/apis/apps/kfconfig/types.go#L62

kunmingg:
@jlewi 
On router side I try to not use disk to avoid race condition among threads, and replace go-getter with http.get. So if we use go-getter on coordinator then kfctl and kfctl-server will diverge in supporting different types of URIs.

kunmingg:
updated

jlewi:
/lgtm
/approve
/hold
on my end looks good thanks!
@gabrielwen any additional comments?

Issue #73, opened 2019-11-04
KFDef v1 - Define API, Conversion Logic, and Update Manifests


jlewi:
In the next release we will want to change KFDef to v1 (from v1beta1).
The API should be largely unchanged.
Nontheless there's some work to be done to actually get all the code in place and update the manifests.

/cc @gabrielwen

gabrielwen:
I think the majority of work is on cleaning weird deps in current codes.  Other than that it should be just adding a new converter.

Issue #68, opened 2019-10-31
KfConfig should be private to coordinator.go


jlewi:
I think KFConfig should be considered an internal private data structure only accessible to the packages in 
https://github.com/kubeflow/kubeflow/tree/master/bootstrap/pkg/kfapp

Code outside pkg/kfapp should not use KFConfig. Instead code outside of pkg/kfapp should use one of the versioned KFDef structures which is the public facing API.

The purpose of KFConfig is to provide an internal datastructure for the code in pkg/kfapp so that code can be written once but work with multiple versions of KFDef.

This is a common pattern when implementing an API; i.e. convert an external versioned data structure into an internal data structure that can evolve independently of the external API.

If code outside of kfapp uses KFConfig then it is no longer an internal datastructure it is a public API that other libraries could end up depending on.

gabrielwen:
agree that we should move `KfConfig` to `/kfapp` only.  I think the structure in overall should be:
```
/pkg:
  - /apis/...
    - /kfdef
      - /v1alpha1
      - /v1beta1
      - /v1
      ...
    - /plugins
      - /gcp
        - /v1alpha1
        - /v1beta1
        ...
      - /aws
      ...
  - /kfapp
    - /kfconfig
      - /converters
    ...
```

but for some similar fields, I'm not sure if we should have reference/dependencies.  IMO I definitely don't want fields being shared between external API (`KfDef` and plugins) and internal data structure (`KfConfig`).  As for shared fields between `KfDef` and plugins, I think it'll be reasonable if plugins are also versioned.  Otherwise it'll be weird and hard to maintain to have a rolling changed data structure to depend on something that is snapshoted and versioned.  So I guess my only question is: ***Do we want to have versioned plugin API as well?***

jlewi:
I think it would make sense for plugin specs to be versioned as well but should the version be kept in sync KFDef?

What should we do in this PR?

gabrielwen:
I don't think they have to be in sync.

Probably not, let's try to roll this out incrementally...  Will duplicate plugins to the new places gradually and change the deps afterwards.

gabrielwen:
`KfConfig` has bee moved to `/pkg/kfconfig`, will close this after we removed the one in `apis/apps`

Jeffwan:
@jlewi @gabrielwen 

I may miss some context and I am a little bit confused now, we have plugins defined at 

[...]

PR #78, opened 2019-11-06
Fix deps in kfapp and kfdef

gabrielwen:
Related: #73 #68 
This is part of removing deps on `kfdef` in `/kfapps` folders.  This PR makes plugins both `/gcp` and `/aws` self-contained so that their dependencies on `SecretRef` and `RepoRef` are removed.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/78)


jlewi:
@gabrielwen Could you please update the PR description with a bit more explanation on what this change is doing, why & how?

gabrielwen:
> @gabrielwen Could you please update the PR description with a bit more explanation on what this change is doing, why & how?

updated.

gabrielwen:
since you are removing `UseIstio` field, do you mind also remove it here?
https://github.com/kubeflow/kfctl/blob/08a2e3736b2ca25f1f7a327090ecc48e014aa90b/pkg/apis/apps/kfconfig/types.go#L62

gabrielwen:
@jlewi can we merge this?  I've had another PR working on #68

PR #84, opened 2019-11-07
Copy plugins from kfapp to apis/apps as external API

gabrielwen:
Related: #68 

Plugins should be part of external facing API.  Moving it to the appropriate place.

For plugins and KFDef we want to have two sets of structs
  
* The external versioned structs (to be used by users and external client libraries)
* The internal structs (to be used internally by kfctl libraries)

This makes it easier to evolve the external facing APIs because if the external APis change we can just write appropriate conversion routines to the internal data structures and not have to update all of the code that is using the internal structs.

 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/84)


gabrielwen:
@jlewi this PR is ready :)

gabrielwen:
@jlewi @kunmingg could I get another quick lgtm?

PR #85, opened 2019-11-07
Copy KfConfig to /pkg

gabrielwen:
Related: #68 

Copy `KfConfig` to `/pkg`.  Didn't copy to `/pkg/kfapp` is because `/kfupgrade` also uses it.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/85)


jlewi:
@gabrielwen Are we copying KfConfig or moving it? If we are copying it then what happens the current version of KFConfig?

In this PR or another PR it might be good to add to the README explaining the directory structure and in particular the difference between public and internal datastructures

gabrielwen:
> @gabrielwen Are we copying KfConfig or moving it? If we are copying it then what happens the current version of KFConfig?
> 
> In this PR or another PR it might be good to add to the README explaining the directory structure and in particular the difference between public and internal datastructures

It's copy in this PR and noop in actual binary, #93 switches to this new file and deletes the old one.  Will update the description in #93

gabrielwen:
/hold
adding internal structure for plugins.

kunmingg:
/lgtm 
/hold
@jlewi any more comments?

jlewi:
@gabrielwen so this PR is putting all the internal structs under /kfconfig; i.e. we are moving the plugin structures into that package.

In the follow on PR #93; the implementation of the plugins will be updated to use the kfconfig structures.

/lgtm
/approve
/hold cancel

PR #103, opened 2019-11-21
adds kfctlServer

swiftdiaries:
Why do we need this?
- The kfctl proto can be used to create servers and clients with any language from this

Use cases:
- Downstream UI apps to deploy Kubeflow ex. UI app using [grpc-web](https://github.com/grpc/grpc-web)
- Downstream language Servers and Clients / SDKs like the sample client written in Go.

/cc @kunmingg @jlewi @gabrielwen

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/103)


jlewi:
We already have a kfctl server.
https://github.com/kubeflow/kubeflow/blob/master/bootstrap/cmd/bootstrap/app/kfctlServer.go

swiftdiaries:
Yes, we do have a server already.
The use case was for cross-language support for installing Kubeflow.
I'm going to follow this PR with a PR in fairing so that the expected UX is:

`pip install kubeflow-fairing`

```
from kubeflow import fairing
fairing.config.set_kfdef("gcp_basic_auth.yaml")
fairing.config.init()
```

This is expected to be used [here](https://github.com/CiscoDevNet/MLAnywhere/blob/master/mla_app_code/mla_core_code.py#L412). Currently, they are shelling out to `kfctl` to install and they're using an older version of Kubeflow. This would enable them to use the server and client associated with the proto more organically.

jlewi:
* If you need a server why can't we just reuse the existing server?
* Why are you defining a new public ApplyKFDef function?
* Why proto why not just REST?

The current server API is basically defined here
https://github.com/kubeflow/kubeflow/blob/183919dccbe1e5c77aece17fd5064992383dddb6/bootstrap/cmd/bootstrap/app/kfctlServer.go#L50

We are basically following the standard Kubernetes CRUD API. Currently we have Create and Get methods which take return KFDef methods. 

Is that API sufficient?

jlewi:
Regarding the fairing bit

```
from kubeflow import fairing
fairing.config.set_kfdef("gcp_basic_auth.yaml")
fairing.config.init()
```

* Why would fairing be involved in deploying/managing Kubeflow? fairing is focused more on consuming Kubeflow resources
* Given we have a pretty standard CRUD API for deploying Kubeflow; what logic would actually reside in the client library? Can't you just use an http library like requests to call create & get?

Issue #105, opened 2019-11-23
setup kfctl nightly builds


swiftdiaries:
It would be great if we could setup nightly builds for kfctl. It would go well along with the E2E tests and help in development. 

/cc @kkasravi @jlewi @yanniszark

jlewi:
@swiftdiaries I've moved the issue into the kubeflow/kfctl repo. Could you provide an update the current status? Do we have a PR? If so what work remains after that PR? What's blocking that PR if anything?

swiftdiaries:
Sorry for the delay. Opened to PR to close this issue. I'll work through the PR and have it done ASAP

swiftdiaries:
After this PR, we need a nightly periodic prow job defined for this repo so that we have nightly builds associated with that run

jlewi:
@swiftdiaries any update on this?

swiftdiaries:
I had a small question on this, have we moved the ProwJobs to [oss-test-infra](https://github.com/GoogleCloudPlatform/oss-test-infra)? Or would it be ok if we push changes to the ProwJobs to `k8s/test-infra`?

jlewi:
@swiftdiaries we haven't moved our prow jobs yet so if you need to update jobs it will have to be in k8s/test-infra.

What changes need to be made to the prow jobs?

What is the remaining work to close this out?

PR #106, opened 2019-11-23
kfctl binary push to GCS

swiftdiaries:
Addresses issue: https://github.com/kubeflow/kfctl/issues/105

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/106)

### DIFF
--- a/py/kubeflow/kfctl/testing/util/kfctl_go_test_utils.py
+++ b/py/kubeflow/kfctl/testing/util/kfctl_go_test_utils.py
@@ -289,8 +289,9 @@ def kfctl_deploy_kubeflow(app_path, project, use_basic_auth, use_istio, config_p
   return app_path
 
 def push_kfctl_to_gcs(kfctl_path):
-  bucket = "PROW_RESULTS_BUCKET"
-  gcs_path = prow_artifacts.get_gcs_dir(bucket) + "artifacts/kfctl/"
+  bucket = "kubernetes-jenkins"
+  logging.info("Bucket name: %s", prow_artifacts.get_gcs_dir(bucket))
+  gcs_path = os.path.join(prow_artifacts.get_gcs_dir(bucket) + "artifacts/kfctl/")
   util.upload_to_gcs(kfctl_path, gcs_path)
 
 def apply_kubeflow(kfctl_path, app_path):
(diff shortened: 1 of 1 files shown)


py/kubeflow/kfctl/testing/util/kfctl_go_test_utils.py
```diff
@@ -285,6 +289,16 @@ def kfctl_deploy_kubeflow(app_path, project, use_basic_auth, use_istio, config_p
     apply_kubeflow(kfctl_path, app_path)
   return app_path
 
+def push_kfctl_to_gcs(app_path):
+  bucket = "kubernetes-jenkins"
```
jlewi: Why are you hardcoding the bucket here? How are we doing it for other artifacts?

jlewi:
@swiftdiaries overall looks good; do you know why the tests are failing?

py/kubeflow/kfctl/testing/util/kfctl_go_test_utils.py
swiftdiaries: I was following the same logic in the function to see if I can write to the bucket ok. I'll switch it out with these [two functions](https://github.com/kubeflow/testing/blob/master/py/kubeflow/testing/util.py#L845).

swiftdiaries:
I think I'm not passing the bucket in the right format. Let me switch it out with the functions in the `kubeflow/testing` repo that should handle the parsing

swiftdiaries:
I seem to be hitting quota issues. I can see the built kfctl binary [here](https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/kubeflow_kfctl/106/kubeflow-kfctl-presubmit/1199055022751485957/artifacts/build_bin/). I checked it with a linux box and it seems to work fine

jlewi:
Do we have an issue tracking the quota issues? I think @gabrielwen was working on fixing those?

py/kubeflow/kfctl/testing/util/kfctl_go_test_utils.py
jlewi: nit: Maybe use the constant defined here: https://github.com/kubeflow/testing/blob/25b87a36f03c207f69abbc9e18500e9ac75647d0/py/kubeflow/testing/prow_artifacts.py#L19

jlewi:
/hold
Please see my one nit-picky comment.

/lgtm
/approve

py/kubeflow/kfctl/testing/util/kfctl_go_test_utils.py
swiftdiaries: Done

PR #109, opened 2019-11-25
Create README.md for KfDef v1beta1

gabrielwen:
We should provide some context for those fields so that new users could know what they mean.

Should we think about auto-generating the docs?

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/109)


jlewi:
@gabrielwen why was kubeflow-kfctl-postsubmit triggered? Is there something wrong with our prow config?

/lgtm
/approve

It would be great to auto-generate API reference docs for KFDef. I think we do this for some of our other resources; e.g.
https://www.kubeflow.org/docs/reference/tfjob/v1/tensorflow/

Perhaps we can reuse the same tooling?

Issue #102, opened 2019-11-20
Check context when applying configs


gabrielwen:
When doing `kfctl delete`, we should check context in `.kube/config` points to the deployment that matches `KfDef`.  If no match, we should reject `kfctl delete`

PR #110, opened 2019-11-26
Check host before actually delete kf

gabrielwen:
Related: #102

- Logs ClusterName after `Apply`
- Checks ClusterName before `Delete` in kustomize.
- Adds force-deletion flag to `kfctl delete`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/110)

### DIFF
--- a/cmd/kfctl/cmd/delete.go
+++ b/cmd/kfctl/cmd/delete.go
@@ -108,6 +108,13 @@ func setAnnotations(configPath string, annotations map[string]string) error {
 	if err != nil {
 		return err
 	}
-	config.SetAnnotations(annotations)
+	anns := config.GetAnnotations()
+	if anns == nil {
+		anns = map[string]string{}
+	}
+	for ann, val := range annotations {
+		anns[ann] = val
+	}
+	config.SetAnnotations(anns)
 	return kfloaders.WriteConfigToFile(*config)
 }
--- a/cmd/kfctl/cmd/delete.go
+++ b/cmd/kfctl/cmd/delete.go
@@ -55,7 +55,6 @@ var deleteCmd = &cobra.Command{
 			forceDeleteAnn: annValue,
 		})
 
-		// TODO: should we check if the configFilePath is local?
 		kfApp, err = coordinator.NewLoadKfAppFromURI(configFilePath)
 		if err != nil || kfApp == nil {
 			return fmt.Errorf("error loading kfapp: %v", err)
--- a/cmd/kfctl/cmd/delete.go
+++ b/cmd/kfctl/cmd/delete.go
@@ -45,11 +45,7 @@ var deleteCmd = &cobra.Command{
 			return fmt.Errorf("Must pass in -f configFile")
 		}
 
-		kfApp, err = coordinator.NewLoadKfAppFromURI(configFilePath)
-		if err != nil || kfApp == nil {
-			return fmt.Errorf("error loading kfapp: %v", err)
-		}
-
+		// Writes annotations to pass information to kfapps.
 		forceDeleteAnn := strings.Join([]string{kfutils.KfDefAnnotation, kfutils.ForceDelete}, "/")
 		annValue := "false"
 		if deleteCfg.GetBool(string(kftypes.FORCE_DELETION)) == true {
--- a/cmd/kfctl/cmd/delete.go
+++ b/cmd/kfctl/cmd/delete.go
@@ -16,9 +16,12 @@ package cmd
 
 import (
 	"fmt"
+	"strings"
 
 	kftypes "github.com/kubeflow/kfctl/v3/pkg/apis/apps"
 	"github.com/kubeflow/kfctl/v3/pkg/kfapp/coordinator"
+	kfloaders "github.com/kubeflow/kfctl/v3/pkg/kfconfig/loaders"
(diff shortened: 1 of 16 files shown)


jlewi:
@gabrielwen Does ClusterIP belong in the status or should we add it to spec? Should we add it as an annotation?  

Should users be allowed to specify cluster IP? This seems like it would diverge form the `kubectl` pattern; i.e. K8s resource specs don't specify the cluster to run on; that is specified by the context in which kubectl is run.

What are the implications of adding it to spec or status with regards to backward compatibility between v1 and v1beta1?

gabrielwen:
> @gabrielwen Does ClusterIP belong in the status or should we add it to spec? Should we add it as an annotation?
> 
> Should users be allowed to specify cluster IP? This seems like it would diverge form the `kubectl` pattern; i.e. K8s resource specs don't specify the cluster to run on; that is specified by the context in which kubectl is run.
> 
> What are the implications of adding it to spec or status with regards to backward compatibility between v1 and v1beta1?

IMO `kfctl` is not 100% complying to `kubectl` semantics - in `KfDef` we do specify the cluster to run on by filling in app name.

adding a check that if `KfDef` doesn't have ClusterIP we skip the check for backward compatibility.

jlewi:
> adding a check that if KfDef doesn't have ClusterIP we skip the check for backward compatibility.
What happens though if someone tries to use an earlier version of kfctl with the newer version of KFDef that contains ClusterIP? Would we end up removing the field when we serialized the updated spec.

jlewi:
@gabrielwen it looks like object metadata includes a clusterName field.
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#objectmeta-v1-meta
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.14/#objectmeta-v1-meta

Should we just use that field in the KFDef metadata?

gabrielwen:
> @gabrielwen it looks like object metadata includes a clusterName field.
> https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#objectmeta-v1-meta
> https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.14/#objectmeta-v1-meta
> 
> Should we just use that field in the KFDef metadata?

The reason I didn't go with clusterName option was because the context name might be different:
- `kfctl` creates a context within kubeconfig using clusterName.
- if user uses `gcloud get-credentials` they'll have context name in some format `gke_...`
- not sure about other tools.

this is why I think using host is a better idea.

let me take a look how to use annotation, tho.

to the question about not having the annotation during delete, I was thinking it's easier that we still issue delete with warning message.  but adding a force flag might be a good idea.

gabrielwen:
@jlewi update the PR.

for the force-deletion, I added a flag in `kfctl delete` and pass this value to kfapps by using annotations.  do you think this is a good idea?

jlewi:
@gabrielwen 

> The reason I didn't go with clusterName option was because the context name might be different:

Why do we need to map clusterName to context name? 

It looks like .kubeconfig maintains a list of clusters each of which has a name, and then contexts reference clusters by name.

On GCP the cluster names seem to follow the convention

```
gke_${PROJECT}_${ZONE}_${CLUSTER_NAME}
```

So the cluster name should be well specified; i.e. if we were to store this name in clusterName it maps to a unique cluster and doesn't rely on the context name which would be specific to a user's .kubeconfig file.

So given the above we cluster name we can

1. Determine that it is a GKE cluster
1. Search for a kubecontext that maps to that cluster
1. If no such context exists run get-credentials to get the appropriate credentials.

gabrielwen:
> @gabrielwen
> 
> > The reason I didn't go with clusterName option was because the context name might be different:
> 
> Why do assume that clusterName is the same as context name?
> 
> It looks like .kubeconfig maintains a list of clusters each of which has a name, and then contexts reference clusters by name.
> 
> On GCP the cluster names seem to follow the convention
> 
> ```
> gke_${PROJECT}_${ZONE}_${CLUSTER_NAME}
> ```
> 
> So the cluster name should be well specified; i.e. if we were to store this name in clusterName it maps to a unique cluster and doesn't rely on the context name which would be specific to a user's .kubeconfig file.
> 
> So given the above we cluster name we can
> 
> 1. Determine that it is a GKE cluster
> 2. Search for a kubecontext that maps to that cluster
> 3. If no such context exists run get-credentials to get the appropriate credentials.

@jlewi 
yea I know about the naming convention and I think this only works on GKE.  I prefer using host is because it should also works on any k8s clusters.

jlewi:
@gabrielwen My expectation would be the same pattern would true on other clouds. Suppose we do the following

If clusterName isn't set

* Fetch the current context
* Get the name of the cluster and store that in clusterName

If clusterName is set

* Check that it matches the name of the cluster with the current context

That seems like a portable solution 

My expectation would be that other cloud providers would (or should) follow a similar pattern in terms of encoding salient information in the clusterName. e.g. for EKS I'd expect something like

```
aws_${ACCOUNT}_${REGION}_${CLUSTER}
```

So that similar to what we will do for GKE, for EKS kfctl could obtain appropriate credentials and create an appropriate KubeConfig entry given the clustername.

@Jeffwan  WDYT?

@gabrielwen it might be worth bringing this up at the community meeting.

/assign @Jeffwan

gabrielwen:
@jlewi updated, please take a look when you have chance.

PR #112, opened 2019-11-27
Set depth=all when checking out a specific branch

jlewi:
* If we need a specific branch then we can't clone with depth=2 we need to use
  depth=all; otherwise the checkout of the branch will fail

* The periodic tests for the v0.7-branch are currently failing because the
  checkout is failing because we can't checkout the requested
  branch kubeflow/testing#498

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/112)


jlewi:
unittests failed trying to delete the directory I suspect it was related to recent issues with our NFS share.

/test all

PR #117, opened 2019-12-02
Kfdef v1

gabrielwen:
Related: #73

Adding `KfDef` v1 - mostly copied from v1beta1.

Please let me know if you think we need to update the definitions!

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/117)


kunmingg:
Shall we take care this: https://github.com/kubeflow/kfctl/blob/master/pkg/apis/apps/kfconfig/types.go#L53

gabrielwen:
> Shall we take care this: https://github.com/kubeflow/kfctl/blob/master/pkg/apis/apps/kfconfig/types.go#L53

that's `KfConfig`.  but yea we should do that as internal code improvement.

jlewi:
> Please let me know if you think we need to update the definitions!

What does update definitions mean? 

Are there any changes to the KFDef spec relative to v1beta1.

/lgtm
/approve
/hold

Assuming there are no API changes relative to v1beta1 you can go ahead and remove the hold.

PR #136, opened 2019-12-11
disable dex test

gabrielwen:
Related: kubeflow/testing#542

Disable the test until we could verify it runs again.

@yanniszark could you help update the test to use the new semantics - write pytest suite and use pyfunc to create the workflow?  you could check how gcp_iap test does it.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/136)

### DIFF
--- a/prow_config.yaml
+++ b/prow_config.yaml
@@ -59,30 +59,30 @@ workflows:
       test_endpoint: true
       config_path: https://raw.githubusercontent.com/kubeflow/manifests/master/kfdef/kfctl_gcp_basic_auth.yaml
   # E2E tests for kfctl_istio_dex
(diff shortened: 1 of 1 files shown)


jlewi:
/lgtm
/approve

/hold
@gabrielwen  Could you file a bug about moving the DEX test to kubeflow/kfctl and updating it? Once you file that issue you can remove the hold.

gabrielwen:
> /lgtm
> /approve
> 
> /hold
> @gabrielwen Could you file a bug about moving the DEX test to kubeflow/kfctl and updating it? Once you file that issue you can remove the hold.

filed.
/hold cancel

PR #144, opened 2019-12-16
fix the bug that only recognize iam service account

### LINKED ISSUE #143: gcp.go can't recognize developer.gserviceaccount.com as service account
In [getIapAccount](https://github.com/kubeflow/kfctl/blob/master/pkg/kfapp/gcp/gcp.go#L1116-1121) we cannot correctly identify there are other possible patterns that could also be serviceAccount.  for example:
`999-compute@developer.gserviceaccount.com`

gabrielwen:
Fixes #143

there are cases service account is other type.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/144)

### DIFF
--- a/pkg/kfapp/gcp/gcp.go
+++ b/pkg/kfapp/gcp/gcp.go
@@ -1112,10 +1112,10 @@ func (gcp *Gcp) copyFile(source string, dest string) error {
 	return nil
 }
 
-// Helper function to generate account field for IAP.
-func (gcp *Gcp) getIapAccount() string {
+// Helper function to generate account field.
+func (gcp *Gcp) getAccount() string {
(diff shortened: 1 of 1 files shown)


gabrielwen:
/hold
@jlewi do you think it's safe that we open the door to other service accounts than `iam.gserviceaccount.com`?  e.g. the one jeff raised: `developer.gserviceaccount.com`

jlewi:
@gabrielwen supporting other service accounts seems reasonable to me. Service accounts of the form

```
${PROJECTNUMBER}-compute@developer.gserviceaccount.com
```

Corresponds to the default compute engine service account.

The go code is merely trying to determine whether its a user account or service account to determine what the appropriate prefix is in the IAM role. So I don't see any reason not to support all types of service accounts.

/hold cancel

PR #153, opened 2019-12-18
Basic auth endpoint fix

gabrielwen:
witnessed periodic test failure due to SSL handshake error in endpoint test.  making it retry so that we don't fail on flaky calls.  also, DNS name can be too long so make it shorter for that.

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/153)

### DIFF
--- a/py/kubeflow/kfctl/testing/util/gcp_util.py
+++ b/py/kubeflow/kfctl/testing/util/gcp_util.py
@@ -123,32 +123,28 @@ def iap_is_ready(url, wait_min=15):
   return False
 
 def _send_req(wait_sec, url, req_gen):
-  _req_num = 0
-
-  @retry(stop_max_delay=wait_sec * 1000)
-  def _send(url, req_gen, req_num):
+  @retry(stop_max_delay=wait_sec * 1000, wait_fixed=10 * 1000)
+  def _send(url, req_gen):
     resp = None
-    req_num += 1
-    logging.info("sending requests to %s; request number: %s" % (url, req_num))
+    logging.info("sending requests to %s" % url)
     try:
       resp = req_gen()
     except SSLError as e:
-      logging.warning("%s: Endpoint SSL handshake error: %s; request number: %s" % (url, e, req_num))
+      logging.warning("%s: Endpoint SSL handshake error: %s" % (url, e))
     except ReqConnectionError:
       logging.info(
-          "%s: Endpoint not ready, request number: %s" % (url, req_num))
+          "%s: Endpoint not ready" % url)
     if not resp or resp.status_code != 200:
-      logging.info("Basic auth login is not ready, request number %s: %s" % (req_num, get_url))
+      logging.info("Basic auth login is not ready: %s" % get_url)
     else:
       return resp
 
-  return _send(url, req_gen, _req_num)
+  return _send(url, req_gen)
 
 def basic_auth_is_ready(url, username, password, wait_min=15):
   get_url = url + "/kflogin"
   post_url = url + "/apikflogin"
 
-  req_num = 0
   end_time = datetime.datetime.now() + datetime.timedelta(
       minutes=wait_min)
 
--- a/py/kubeflow/kfctl/testing/util/gcp_util.py
+++ b/py/kubeflow/kfctl/testing/util/gcp_util.py
@@ -123,6 +123,18 @@ def iap_is_ready(url, wait_min=15):
   return False
 
 def _send_req(wait_sec, url, req_gen):
+  """ Helper function to send requests and retry when the endpoint is not ready.
+
+  Args:
+  wait_sec: int, max time to wait and retry in seconds.
+  url: str, url to send the request, used only for logging.
+  req_gen: lambda, no parameter function to generate requests.Request for the
+  function to send to the endpoint.
+
+  Returns:
+    requests.Response
+  """
+
   # Simple identifier that we need to re-send the request.
   class RetryError(Exception):
     pass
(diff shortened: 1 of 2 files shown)


py/kubeflow/kfctl/testing/util/gcp_util.py
```diff
@@ -122,46 +122,84 @@ def iap_is_ready(url, wait_min=15):
...
+    logging.info("sending request to %s" % url)
     try:
-      resp = requests.request(
-          "GET",
-          get_url,
-          verify=False)
+      resp = req_gen()
     except SSLError as e:
-      logging.warning("%s: Endpoint SSL handshake error: %s; request number: %s" % (url, e, req_num))
+      logging.warning("%s: Endpoint SSL handshake error: %s" % (url, e))
```
jlewi: Why convert SSLError into RetryError? Why not just have the retry_on_exception test if its an SSLError or ReqConnectionError?

py/kubeflow/kfctl/testing/util/gcp_util.py
```diff
@@ -122,46 +122,84 @@ def iap_is_ready(url, wait_min=15):
...
+
+def basic_auth_is_ready(url, username, password, wait_min=15):
+  get_url = url + "/kflogin"
+  post_url = url + "/apikflogin"
+
+  end_time = datetime.datetime.now() + datetime.timedelta(
+      minutes=wait_min)
+
+  wait_time = datetime.datetime.now() - end_time
+  while True:
```
jlewi: Won't you keep looping forever unless you get a 200? it looks like the original code was only waiting until end_time.

Do we still need this while loop if we are going to wrap the requests using @retry? Can't we just use @retry to set the max duration to try for?

py/kubeflow/kfctl/testing/util/gcp_util.py
```diff
@@ -122,46 +122,84 @@ def iap_is_ready(url, wait_min=15):
...
+    resp = _send_req(wait_time.seconds, get_url, lambda: requests.request(
```
jlewi: nit: I think the code would be simpler if you just did something like

```

 def retry_if_retry_error(e):
    return isinstance(e, [SSLError, ReqConnectionError])

@retry(..., retry_on_exception=retry_if_retry_error)
def send_request(*args,**kwargs):
  return requests.request(*args, **kwargs)
```

py/kubeflow/kfctl/testing/pytests/endpoint_ready_test.py
```diff
@@ -15,7 +15,7 @@
 
 # There's really no good reason to run test_endpoint during presubmits.
 # We shouldn't need it to feel confident that kfctl is working.
-@pytest.mark.skipif(os.getenv("JOB_TYPE") == "presubmit",
+@pytest.mark.skipif(os.getenv("JOB_TYPE") == "periodic",
```
jlewi: Why don't you want to skip the endpoint test in the presubmits?

gabrielwen: this is a temp change for this function.  will rebase and drop it after I can make sure the function works.

jlewi: Got it thanks.

jlewi:
Thanks @gabrielwen let me know when this is ready for another look.

py/kubeflow/kfctl/testing/util/gcp_util.py
gabrielwen: updated.

py/kubeflow/kfctl/testing/util/gcp_util.py
gabrielwen: updated `_send_req`

py/kubeflow/kfctl/testing/util/gcp_util.py
gabrielwen: so we need to be flexible about the condition to retry:
- some exceptions (SSLError, ReqConnectionError)
- some status code (200)

so I added an optional param to `_send_req` for the purpose.

gabrielwen:
/holde cancel
@jlewi updated, thanks!

py/kubeflow/kfctl/testing/util/gcp_util.py
```diff
@@ -122,46 +122,77 @@ def iap_is_ready(url, wait_min=15):
...
+  wait_time = datetime.datetime.now() - end_time
+  resp = _send_req(wait_time.seconds, get_url, lambda: requests.request(
+      "GET",
+      get_url,
+      verify=False), retry_result_code=200)
+
+  logging.info("%s: endpoint is ready; response: %s" % (get_url, resp.text))
+  logging.info("%s: testing login API" % post_url)
+
+  wait_time = datetime.datetime.now() - end_time
```
jlewi: Why not just use wait_min? It looks you are using wait_min to compute end_time which is then used to compute wait_time.  I think you could just pass wait_min * 60 to _send_req

jlewi:
One minor cleanup and then this is ready to go.

py/kubeflow/kfctl/testing/util/gcp_util.py
gabrielwen: won't this elapse the testing time?  say that we have 3 calls to _send_req, the max testing time will become 45 minutes instead of 15 minutes?

jlewi: I see; so you are adjusting the wait time based on the total timeout for the test. That makes sense. Thanks.

PR #165, opened 2020-01-08
update profile test dir in e2e test

kunmingg:
following https://github.com/kubeflow/kubeflow/pull/4633

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/165)

### DIFF
--- a/py/kubeflow/kfctl/testing/ci/kfctl_e2e_workflow.py
+++ b/py/kubeflow/kfctl/testing/ci/kfctl_e2e_workflow.py
@@ -336,7 +336,6 @@ class Builder:
     for t in workflow["spec"]["templates"]:
       if t["name"] == name:
         return t
-    workflow["spec"]["templates"].append(new_template)
 
     return None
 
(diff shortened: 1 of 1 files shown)


py/kubeflow/kfctl/testing/ci/kfctl_e2e_workflow.py
```diff
@@ -336,7 +336,6 @@ def _build_step(self, name, workflow, dag_name, task_template,
     for t in workflow["spec"]["templates"]:
       if t["name"] == name:
         return t
-    workflow["spec"]["templates"].append(new_template)
```
jlewi: Why did this line get deleted?

kunmingg: Seems `new_template ` is not defined anywhere.

jlewi: Thanks. It looks like this is legacy code and this is now handled by 
https://github.com/kubeflow/testing/blob/f5aaa62db495a78816ac31a9673c3a0ba0602c8f/py/kubeflow/testing/argo_build_util.py#L151

Can we delete everything after line 333 and just do

```
return argo_build_util.add_task_to_dag(....)
```

kunmingg: updated

Issue #167, opened 2020-01-09
kfctl throws EOF error


kunmingg:
Error from kfctl build:
```
Code 500 with message: could not sync cache. Error:  (kubeflow.error): Code 400 with message: couldn't download URI https://github.com/kubeflow/manifests/archive/master.tar.gz Error Error opening a gzip reader for /var/folders/7c/p1bjgh7s4nb1lmt4q9y2258h00hrhl/T/getter189618912/archive: EOF
```
How to reproduce:
```
kfctl build -f https://raw.githubusercontent.com/kubeflow/manifests/master/kfdef/kfctl_gcp_iap.yaml
```

Anderssorby:
I think the reason is that kfctl doesn't follow redirects and github has changed their setup to use the subdomain codeloads.github.com. Manually changing the URIs in the config to the redirected URLs seems to fix the issue.

kunmingg:
fix included in https://github.com/kubeflow/kfctl/tree/v0.7.2-rc.0

heoelri:
I'm facing a simmilar problem with Kfctl 0.5.1. Is it planned rollout the fix to older versions of kfctl?

```bash
wget https://github.com/kubeflow/kubeflow/releases/download/v0.5.1/kfctl_v0.5.1_linux.tar.gz -O kfctl.tar.gz
tar -xvf kfctl.tar.gz
KFPWD=`pwd`; export PATH=$PATH:$KFPWD
export KFAPP="kubeflow"
kfctl init ${KFAPP}
cd ${KFAPP}
```

```console
FATA[0000] could not download repo to cache Error  (kubeflow.error): Code 400 with message: couldn't download kubeflow repo https://github.com/kubeflow/kubeflow/tarball/v0.5.1?archive=tar.gz Error Error opening a gzip reader for /tmp/getter924981679/archive: EOF  filename="coordinator/coordinator.go:304"
```

maximmold:
To clarify, you would update the repos section of your config file:

From this
```
  repos:
  - name: manifests
    uri: https://github.com/kubeflow/manifests/archive/v0.7-branch.tar.gz
```

To this

```
  repos:
  - name: manifests
    uri: https://codeload.github.com/kubeflow/manifests/tar.gz/v0.7-branch
```

PR #169, opened 2020-01-10
fix kfctl

kunmingg:
kfctl was broken: https://github.com/kubeflow/kfctl/issues/167

replace `gogetter` with http get

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/169)

### DIFF
--- a/pkg/kfconfig/types.go
+++ b/pkg/kfconfig/types.go
@@ -21,6 +21,7 @@ import (
 	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"github.com/otiai10/copy"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


ryandawsonuk:
I hit this error so tried to install from this fork. It does resolve the EOF error and I get much further.

Eventually though I hit this:
```
failed to apply:  (kubeflow.error): Code 500 with message: kfApp Apply failed for kustomize:  (kubeflow.error): Code 500 with message: error evaluating kustomization manifest for istio Error absolute path error in 'kustomize/istio' : evalsymlink failure on 'kustomize/istio'
```
I am using https://github.com/ryandawsonuk/seldon-gitops/blob/master/kfctl_istio_dex_noseldon.yaml instead of https://github.com/kubeflow/manifests/blob/master/kfdef/kfctl_istio_dex.yaml but I've compared the yaml on them and the differences are minimal.

ryandawsonuk:
Ok that one must be my problem as I don't get that if I do `kfctl apply -f https://raw.githubusercontent.com/kubeflow/manifests/master/kfdef/kfctl_istio_dex.yaml -V`

I do get `"metadata-ui" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app":"metadata-ui", "app.kubernetes.io/component":"metadata", "app.kubernetes.io/instance":"metadata-v0.7.0", "app.kubernetes.io/managed-by":"kfctl", "app.kubernetes.io/name":"metadata", "app.kubernetes.io/part-of":"kubeflow", "app.kubernetes.io/version":"v0.7.0", "kustomize.component":"metadata"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable` if I run on an existing cluster. But I guess that's a different issue.

pkg/kfconfig/types.go
```diff
@@ -530,6 +551,54 @@ func (c *KfConfig) SyncCache() error {
 	return nil
 }
 
+func untar(body []byte, cacheDir string) error {
```
abhi-g: golang has an untar library. Can this be used instead or reimplementing the function..
https://godoc.org/golang.org/x/build/internal/untar

PR #183, opened 2020-01-16
Update dependency to remove bou.ke/monkey

zhenghuiwang:
Run `go mod tidy` to remove unused packages.

One is `	github.com/otiai10/curr` which includes ` bou.ke/monkey`.

 bou.ke/monkey doesn't give any permission to use it:

https://github.com/bouk/monkey/blob/master/LICENSE.md

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/183)


abhi-g:
So no code change was required, ie, it was a transitive dependency that wasn't being used?

kunmingg:
Yes, only remove unused import in go.mod

PR #185, opened 2020-01-21
fix third party license download links

adrian555:
Some download links for third party licenses are not downloading the raw contents, therefore, `make test` fails. This fixes the download links and `make test`.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/185)


adrian555:
Close this one since the fix is included in PR #186.

PR #187, opened 2020-01-21
Fix kustomize secretGenerator merging bug

### LINKED ISSUE #156: kustomize: secretGenerator merge not handled correctly
After applying the db overlay for the metadata component of kubeflow from manifests, I noticed that a secret was not generated.
The merged kustomization file was missing the secretGenerator field.
I suspect this is where the faulty logic is:
https://github.com/kubeflow/kfctl/blob/9957ea14f901f3de107c059561e1de0a8529b497/pkg/kfapp/kustomize/kustomize.go#L752-L764

I believe the issue is not handling the case where a merge behavior is not specified.

/kind bug
/priority p1
/cc @gabrielwen @kkasravi @jlewi

Tomcli:
The new metadata in manifests now using the secretGenerator for creating some of its secrets. However, kfctl's secretGenerator is missing some fallback condition which causes issue #156. This PR included the following changes: 
- Add default switch case when none of the secretGenerator is matched.
- Convert part of the updateConfigMapArgs into a common helper function for changing env path on both SecretArgs and ConfigMapArgs.

Fixes #156

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/187)

### DIFF
--- a/pkg/kfapp/kustomize/kustomize.go
+++ b/pkg/kfapp/kustomize/kustomize.go
@@ -20,6 +20,15 @@ import (
 	"bufio"
 	"encoding/hex"
 	"fmt"
+	"io/ioutil"
+	"math/rand"
+	"os"
+	"path"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"time"
+
 	"github.com/cenkalti/backoff"
 	"github.com/ghodss/yaml"
 	"github.com/imdario/mergo"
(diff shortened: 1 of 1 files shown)


Tomcli:
Looks like the presubmit test failed on some of the gcp auths/connections.

Tomcli:
kfam failed again
/retest

zhenghuiwang:
/cc @kunmingg @jlewi 

Could you help review this fix for https://github.com/kubeflow/manifests/issues/799 ?

zhenghuiwang:
@Tomcli can you rebase this PR so that it can pass the presubmit tests? Thanks

Tomcli:
@zhenghuiwang @kunmingg Thanks, I rebased my commits.

PR #197, opened 2020-01-30
Add a link to operator docs to README

vpavlin:
As there is no `docs/` directory and operator documentation is in root dir, it would be good to have a link in README

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/197)


animeshsingh:
@vpavlin am doing a larger updates to the docs here, and linking to operators is part of that. Will address it there

PR #198, opened 2020-01-30
Fix katib deployment tests

richardsliu:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/198)


jlewi:
@richardsliu rather than comment it out just mark the test as expected to fail using the xfail pytest macro.

http://doc.pytest.org/en/latest/skipping.html

This way we can still get some signal without blocking the tests.

jlewi:
/lgtm
/approve

It looks like test_deploy.py might no longer be used so we might be able to just delete it.

PR #221, opened 2020-02-05
bump kustomize dependency to v3.2.0 (#220)

### LINKED ISSUE #194: kfctl set-image-name command appends inventory:omitempty: null into all the kustomization.yaml files
kfctl set-image-name command appends "inventory:omitempty: null" into all the kustomization.yaml files in the manifests dir

Command to reproduce the issue:
kfctl set-image-name test-repo.io

with this "inventory:omitempty: null" key-value pair in the yaml file, kfctl apply fails.

Workaround by deleting the added line.
                                        re := regexp.MustCompile("(?m)[\r\n]+^.*omitempty.*$")
                                        res := re.ReplaceAllString(string(data), "")
                                        ........
                                        ........

                                        if _, err := file.Write([]byte(res)); err != nil {
						return err
					}

xaniasd:
Resolves #194

There's a bug in v3.1.0 [here](https://github.com/kubernetes-sigs/kustomize/blob/v3.1.0/pkg/types/kustomization.go#L142) that introduces a faulty entry when updating the kustomization file with set-image-name.

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/221)


xaniasd:
@jlewi is this what you meant?

em-pe:
Hi @xaniasd thanks for the fix! @jlewi could someone have a look at this review? The fix is already on master but for 1.0 rc `set-image-name` is still unusable.

Jeffwan:
This is a breaking change and kubeflow/manifest still use 2.x syntax and need to be updated as well. I think we can revisit this improvement after 1.0 release.

xaniasd:
hi @Jeffwan, good point. On the other hand, at least the tests in kubeflow/manifests must be using 3.x syntax, see [code](https://github.com/kubeflow/manifests/blob/98b380d6081dfdcf2861abbc64f74cb22d02a8a4/go.mod#L5) and kubeflow/manifests#309

In my opinion it's better to avoid these inconsistencies and update the rest of the stuff as soon as possible. I'll see if I can make a PR

Jeffwan:
@gabrielwen @jlewi

jlewi:
What does bumping the kustomize version have to do with fixing set-image-name? 

@xaniasd Could you please update the PR description to explain what this is fixing and why we need it on the 1.0 branch?

xaniasd:
hi @jlewi, this PR is per your request, see [here](https://github.com/kubeflow/kfctl/issues/194#issuecomment-582099731). Throwback to when I was having my thesis reviewed back in the day :laughing:

I'll edit the description, no worries.

jlewi:
Great thanks.

/lgtm
/approve

katiesandford:
@jlewi would it be possible to publish a new release for the kfctl binary?
The latest one here (https://github.com/kubeflow/kfctl/releases/) doesn't include this change and so can't be used to change images to use a custom gitlab registry

PR #223, opened 2020-02-07
Add tests to check status metadata deployments

zhenghuiwang:
Add checks for whether metadata deployments start.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/223)


py/kubeflow/kfctl/testing/pytests/kf_is_ready_test.py
```diff
@@ -101,6 +101,8 @@ def test_kf_is_ready(record_xml_attribute, namespace, use_basic_auth, use_istio,
       "centraldashboard",
       "jupyter-web-app-deployment",
       "minio",
+      "metadata-deployment",
```
jlewi: Instead of adding it to the existing test can we create a separate function for it? Similar to test_katib_is_ready.
I think we want to have many small tests as opposed to 1 monolithic test. This makes it easier to deal with flakes;
e.g. if one particular app is flaky we mark that specific test as expected to fail.

In spyglass and testgrid we also know which application is having problems as opposed to just kf isn't ready.

zhenghuiwang: Good suggestion! Done.

py/kubeflow/kfctl/testing/pytests/kf_is_ready_test.py
```diff
@@ -62,16 +61,29 @@ def test_katib_is_ready(record_xml_attribute, namespace):
 
   util.load_kube_config()
 
+  for deployment_name in deployments:
+    logging.info("Verifying that deployment %s started...", deployment_name)
+    util.wait_for_deployment(api_client, namespace, deployment_name, 10)
+
+def test_katib_is_ready(record_xml_attribute, namespace):
```
jlewi: Do you want to add back pytest.mark.xfail?

py/kubeflow/kfctl/testing/pytests/kf_is_ready_test.py
```diff
@@ -62,16 +61,29 @@ def test_katib_is_ready(record_xml_attribute, namespace):
...
     "katib-db-manager",
     "katib-ui",
   ]
-  for deployment_name in deployment_names:
-    logging.info("Verifying that deployment %s started...", deployment_name)
-    util.wait_for_deployment(api_client, namespace, deployment_name, 10)
+  check_deployments_ready(record_xml_attribute, namespace,
+                          "test_katib_is_ready", deployment_names)
 
+def test_metadata_is_ready(record_xml_attribute, namespace):
```
jlewi: @zhenghuiwang while your here; would you mind replacing test_kf_is_ready with several tests for the remaining things in kf_is_ready?

So maybe

* 1 for central dashboard
* 1 for profiles
* 1 for all the pipelienes stuff (including Argo)
* 1 for TFJob
* 1 For PyTorch
* 1 For Jupyter apps?

zhenghuiwang: Done.

py/kubeflow/kfctl/testing/pytests/kf_is_ready_test.py
zhenghuiwang: Verified with @richardsliu that this test should pass now so that i removed the mark.

jlewi:
Thanks @zhenghuiwang 
/lgtm
/approve

PR #225, opened 2020-02-07
Add IAM Role and Policy for Service Account support

Jeffwan:
https://github.com/kubeflow/kubeflow/issues/4403

Expose an option for users to configure IAM roles and policy for service account. 

We will create two IAM roles
- kf-admin-${clusterId}
- kf-user-${clusterId}

they will bind to kubernetes service account, `kf-admin` will be used for control plane AWS service authentication, like ALB, EFS CSI, etc.  `kf-user` and service account under user's namespace will be used to talk to S3 (pipeline artifact), ECR (fairing repo submit), etc.
- kf-admin
- kf-user

Profile controller change will be checked in a separate PR.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/225)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,7 +20,6 @@ require (
 	github.com/go-openapi/jsonpointer v0.19.2 // indirect
 	github.com/gogo/protobuf v1.3.1
 	github.com/google/go-cmp v0.3.1
-	github.com/google/martian v2.1.0+incompatible
 	github.com/gregjones/httpcache v0.0.0-20190611155906-901d90724c79 // indirect
 	github.com/hashicorp/go-getter v1.0.2
 	github.com/imdario/mergo v0.3.8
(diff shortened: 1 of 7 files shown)


jlewi:
/assign @gabrielwen 

@gabrielwen any concerns about cherry-picking this for 1.0? It looks like the code is only affecting AWS

gabrielwen:
/lgtm
/approve

yea looks like it's only on the AWS side.  should be fine for a cherrypick.

devgrok:
Cheers for the work in adding the IAM Role / Service Account support.

A bit late to ask but I have a query:

Wouldn't it be better if the IAM role ARN was set as parameter on each application (i.e. `aws-alb-ingress-controller`) so that kustomize applies the annotation?
This would also fit better with using a CD to apply manifest updates instead of `kfctl apply` and it would also give an audit of what the assignment is. 

This would also make it easier to create the roles prior to running kubeflow/kfctl in the case where they wanted to grant different permissions to those ServiceAccounts (like only granting `iam_alb_ingress_policy` to the `aws-alb-ingress-controller` ServiceAccount)

PR #226, opened 2020-02-10
kfctl image mirror commands

kunmingg:
kubeflow  image mirror: mirror images in `kustomize` config to new registry
- generate tekton & cloud build pipeline to mirror the image
- update image fields in `kustomize` config

**Pipeline Build & Run:**

`kfctl alpha mirror build <config_file> -V -o <pipeline_file> [--gcb]`:
loop through `kustomize` folder under current dir, generate image mirror pipeline.
- config_file: defines mirroring destination registry, build context and affected images. for example [gcr_template](https://github.com/kubeflow/manifests/blob/master/experimental/mirror-images/gcp_template.yaml)
- pipeline_file: file name of output [tekton pipeline](https://github.com/tektoncd/pipeline)
- --gcb: when set, buidl will output cloud build pipeline `cloudbuild.yaml`

We support running image mirroring tasks with cloud build or tekton pipeline.
each task in the pipeline is using Dockerfile https://github.com/kubeflow/kubeflow/pull/4756

1. To run the gcr cloud build pipeline:
`gcloud builds submit gs://kubeflow-examples/image-replicate/replicate-context.tar.gz --project <project_id> --config cloudbuild.yaml`

2. To run the tekton pipeline:
`kubectl apply -f <pipeline_file>`
 
**Kustomize Update:**

`kfctl alpha mirror overwrite -i <pipeline_file>`:
loop through `kustomize` folder under current dir, edit manifests to use new images.
- pipeline_file: input tekton pipeline defining old_image -> new_image mapping

The goal of breaking image mirroring process into 2 cmds is to make it robust, idempotent, and reusable.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/226)


kunmingg:
@jtfogarty

hemantha-kumara:
Isn't this similar to `kfctl set-image` 
At least 1 point where replacing registry in manifests ?

kunmingg:
@hemantha-kumara 
`kfctl set-image` will replace `gcr.io/kubeflow-images-public` to private GCR, while keep other images unchanged.
`kfctl alpha replicate-mages` will replace non-GCR to private GCR, while keep `gcr.io/kubeflow-images-public` images unchanged.

jlewi:
@kunmingg WDYT about the following semantics

```
kfctl alpha build-mirror-pipeline -f patterns.yaml -o pipeline.yaml --manifests=${PATH_TO_KUBEFLOW_MANIFEST}
```

where `patterns.yaml` is a file like the following

```
apiVersion: imagereplace.kubflow.org/v1alpha1
kind: imagereplace
spec:
  patterns
     - source: quay.io/
       dest:  gcr.io/my-project
     - source: docker.io/
       dest: gcr.io/my-project
```

So the logic isn't specific to GCR.

kunmingg:
@jlewi @abhi-g 
updated following offline discussion.

jlewi:
Thanks @kunmingg  to recap

1. Lets make the filenames for input and output explicit flags rather than doing hardcoded names
1. I think you want to emit a PipelineRun and then inline the pipeline into the Pipeline
1. Please include a sample YAML file  in the PR.

kunmingg:
updated

cmd/kfctl/cmd/replicate-build.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/spf13/viper"
+)
+
+var include string
+var exclude string
+var context string
+var outputFileName string
+
+func init() {
+	replicateBuildCmd.Flags().StringVarP(&include, "include", "i", "", "When set, only replicate images with matching prefix")
```
jlewi: Instead of using flags to specify include/exclude patterns, can we define a suitable YAML file which lists the patterns and replacements?

cmd/kfctl/cmd/replicate-build.go
```diff
@@ -0,0 +1,59 @@
...
+	replicateBuildCmd.Flags().StringVarP(&exclude, "exclude", "e", "gcr.io", "Skip replicate images with matching prefix")
+	replicateBuildCmd.Flags().StringVarP(&context, "context", "c", "", "Path to build context, for example gs://...")
```
jlewi: Does this need to build a flag? Could a user just edit the PipelineRun that you output?

config/samples/image-replicate-pipeline.yaml
```diff
@@ -0,0 +1,421 @@
+apiVersion: tekton.dev/v1alpha1
```
jlewi: Could we write a unittest and make this the expected value of the unittest?

pkg/kfapp/gcp/replicate_image.go
```diff
@@ -0,0 +1,204 @@
...
+			return nil
+		}
+
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	taskIns := pipeline.Task{
```
jlewi: Do you need to generate the Task using go? Can't we just provide people with a YAML file in the repo?

jlewi:
Thanks @kunmingg; I still have questions about the CLI semantics. It might be faster if we used the PR conversation to agree on the correct semantics before implementing in code. Could you provide updated example invocations for all the command?

My high level comment is I'd really like to minimize the number of command line flags. Instead we should use configuration files.

kunmingg:
1. `kfctl alpha replicate-build -f config.yaml -o pipeline.yaml`:
output tekton pipeline `pipeline.yaml` file

2. `kfctl alpha replicate-cb -i pipeline.yaml`:
submit cloud build pipeline to replicate

3. kfctl alpha replicate-overwrite -i pipeline.yaml`:
update images in manifests to use new images specified by pipeline.yaml image mapping.

kunmingg:
sample config for "replicate-build":
https://github.com/kubeflow/manifests/pull/907/files#diff-466dec9580c30811f4df84584718e51e

jlewi:
@kunmingg Thanks overall looks good. Could you write a unittest please to verify the generated pipelinerun is what you expect?

For the command syntax should we make the 3 commands `build-spec`, `cb`, `overwrite` subcommands of another comamnd e.g.

```
kfctl alpha mirror build-spec
kfctl alpha mirror cb
kfctl alpha mirror overwrite
```

Before you change the code lets agree on the commands so you only have to change it once

Also instead of having two build commands could we add a flag for the format e.g.

```
kfctl alpha mirror build   <creates Tekton spec>
kfctl alpha mirror build --format=gcb <creates gcp spec>
```

Longer term as we discussed, overwriting the kustomization seems like an unfortunate side effect of the way we have written our kustomize manifests. It would be better if we could just write a kustomization file and use overlays. Can we file a bug for that?

One thing I was thinking about, in principle we could use a Tekton parameter to specify the dest prefix.

e.g.

```
    - name: replicate-2
      params:
      - name: inputImage
        value: vertaaiofficial/modeldb-artifact-store:kubeflow
      - name: outputImage
        value: $(params.dest)/vertaaiofficial/modeldb-artifact-store:kubeflow
      - name: context
        value: $(params.context)
```

Then users in principle wouldn't even need to generate their own tekton pipeline they could just modify the YAML to set their context and dest. Or we could produce a Pipeline and they could then just define a PipelineRun which references that pipeline

This might make the overwrite logic a bit harder because we'd have to combine the parameter with the substitution in order to get the dest values.

WDYT?

Also could we generate more informative step names rather than `replicate-2`? Could we use the name of the image and maybe include the number to ensure uniqueness?

kunmingg:
How about
```
kfctl alpha mirror build   <creates Tekton spec>
kfctl alpha mirror build --format=gcb <creates Tekton and gcp spec>
```
since `kfctl alpha mirror overwrite` will expect an Tekton pipeline as input?

Benefits of generating tekton pipeline dynamically is we don't need to maintain a pipeline release;
and user can mirror all kustomize pkg, not only kubeflow.

jlewi:
@kunmingg LGTM to the proposed semantics. I think you can go ahead and implement that.

> Benefits of generating tekton pipeline dynamically is we don't need to maintain a pipeline release;
and user can mirror all kustomize pkg, not only kubeflow.

Even if we made the base of destination image a Tekton parameter, we could still not publish it and ask users to generate it. 

I'm fine either way; users could always do a find replace to make it a parameter. So if you prefer to not make it a parameter right now that seems good to me. 

I think you can go ahead and implement the new semantics and then we can merge this.

animeshsingh:
thanks @kunmingg - is the goal that this is aware of *ALL* the images for a particular Kubeflow release, can download and populate a local registry target?

animeshsingh:
> `kfctl set-image` will replace `gcr.io/kubeflow-images-public` to private GCR, while keep other images unchanged.
> `kfctl alpha replicate-mages` will replace non-GCR to private GCR, while keep `gcr.io/kubeflow-images-public` images unchanged.

so this cannot replicate image in GCR to local registry?

kunmingg:
When user mirror images with tekton pipeline, any target registry should be fine as long as user have write access to it.

jlewi:
Thanks @kunmingg just a couple nit-picky comments.

kunmingg:
@jlewi 
https://github.com/kubeflow/manifests/pull/914

animeshsingh:
> When user mirror images with tekton pipeline, any destination registry should be fine as long as user have write access to it.

@kunmingg but your statement here says

"kfctl alpha replicate-mages will replace non-GCR to private GCR, while keep gcr.io/kubeflow-images-public images unchanged"

what if i want to replicate all the public images (gcr/dockerhub) to a local docker registry?

kunmingg:
@animeshsingh 
PR description updated. 
This feature is not GCR only.
You can try out after https://github.com/kubeflow/kfctl/pull/239 merged.

PR #239, opened 2020-02-14
mirror overwrite: drop image tags before write back

kunmingg:
fix bug for `mirror overwrite `

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/239)


jlewi:
@kunmingg does the code handle shas as well?

kunmingg:
yes
/hold cancel

Issue #228, opened 2020-02-12
endpoint_is_ready test is failing


gabrielwen:
Seems like to be related to Python 2.x deprecation:
https://stackoverflow.com/questions/47142848/python-sslerror-bad-handshake-unexpected-eof

Updating test worker Dockerfile.

jlewi:
@gabrielwen I just saw this. It looks like we might have both been working on updating the test worker image.

See https://github.com/kubeflow/testing/pull/606

jlewi:
@gabrielwen any update on this?

gabrielwen:
still working on the py3 image.

gabrielwen:
kubeflow/tf-operator#1133 has to be checked in first so that #227 can be fixed.

jlewi:
@gabrielwen could you please explain what the issue is? IIRC there is an issue with the endpoint test requiring python3? 

Do we need to update all the tests to use python3 just to fix the endpoint ready test? Could we just run the endpoint is ready test using the new docker image and let all the remaining tests continue to use the existing python2 based docker image?

gabrielwen:
> @gabrielwen could you please explain what the issue is? IIRC there is an issue with the endpoint test requiring python3?
> 
> Do we need to update all the tests to use python3 just to fix the endpoint ready test? Could we just run the endpoint is ready test using the new docker image and let all the remaining tests continue to use the existing python2 based docker image?

you are right we could just use python3 image on endpoint test.  it's just that I thought we'd like to migrate to python3 images.  however, I think changing to python3 on kfctl e2e test is almost completed.

jlewi:
@gabrielwen we should definitely migrate to python3; however it would be good to get the endpoint test fixed as soon as possible so that get clear signal about whether KF is working or not.

jbottum:
@gabrielwen hey Gabriel - can we please close this issue ?

PatrickXYS:
I think this is resolved by https://github.com/kubeflow/kfctl/pull/245

/close

PR #245, opened 2020-02-18
use py3 img only in endpoint_is_ready

gabrielwen:
fix endpoint_is_ready pytest for python2.x deprecation.  use new test-worker image on python3.

presubmit with endpoint test was passed: https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubeflow_kfctl/245/kubeflow-kfctl-presubmit/1229917734066720768/

Related: #228

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/245)

### DIFF
--- a/py/kubeflow/kfctl/testing/ci/kfctl_e2e_workflow.py
+++ b/py/kubeflow/kfctl/testing/ci/kfctl_e2e_workflow.py
@@ -555,6 +555,7 @@ class Builder(object):
     # Change the workfing directory for all subsequent steps
     task_template["container"]["workingDir"] = os.path.join(
       self.kfctl_pytest_dir)
+    py3_template["container"]["workingDir"] = os.path.join(self.kfctl_pytest_dir)
 
     #**************************************************************************
     # Run build_kfctl and deploy kubeflow
--- a/py/kubeflow/kfctl/testing/ci/kfctl_e2e_workflow.py
+++ b/py/kubeflow/kfctl/testing/ci/kfctl_e2e_workflow.py
@@ -439,8 +439,6 @@ class Builder(object):
   def _build_exit_dag(self):
     """Build the exit handler dag"""
     task_template = self._build_task_template()
-    py3_template = argo_build_util.deep_copy(task_template)
-    py3_template["container"]["image"] = "gcr.io/kubeflow-ci/test-worker-py3:789005d"
 
     #***********************************************************************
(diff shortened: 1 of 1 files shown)


py/kubeflow/kfctl/testing/pytests/endpoint_ready_test.py
```diff
@@ -15,7 +15,7 @@
 
 # There's really no good reason to run test_endpoint during presubmits.
 # We shouldn't need it to feel confident that kfctl is working.
-@pytest.mark.skipif(os.getenv("JOB_TYPE") == "presubmit",
+@pytest.mark.skipif(os.getenv("JOB_TYPE") == "periodic",
```
richardsliu: Why change this?

gabrielwen: to enable the test - will rebase once the fix is confirmed.

py/kubeflow/kfctl/testing/ci/kfctl_e2e_workflow.py
```diff
@@ -639,9 +637,11 @@ def build(self):
                  "--use_basic_auth={0}".format(self.use_basic_auth),
               ]
 
+      py3_template = argo_build_util.deep_copy(task_template)
+      py3_template["container"]["image"] = "gcr.io/kubeflow-ci/test-worker-py3:789005d"
```
jlewi: nit: Should we define a constant at the top of the file so its easy to change?

py/kubeflow/kfctl/testing/pytests/endpoint_ready_test.py
jlewi: @gabrielwen nit might be easier to just comment out the  @pytest.... rather than changing  the value to periodic. This way even if the code gets merged we wouldn't disable it in the periodic tests

py/kubeflow/kfctl/testing/ci/kfctl_e2e_workflow.py
gabrielwen: yea, I can move this up to the top.

gabrielwen:
@jlewi  @richardsliu this is ready.

gabrielwen:
@jlewi sorry need your lgtm again

PR #248, opened 2020-02-19
Update operator instructions for using the default kubeflow kfdef

Tomcli:
The default Kubeflow kfdef under kubeflow/manifests now removed all the metadata.name field. Thus, we need to update the operator instructions for those corresponding changes.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/248)

### DIFF
--- a/operator.md
+++ b/operator.md
@@ -24,18 +24,30 @@ kubectl create ns ${KUBEFLOW_NAMESPACE}
 # kubectl create -f deploy/crds/kfdef_quota.yaml -n ${KUBEFLOW_NAMESPACE} # only deploy this if the k8s cluster is 1.15+ and has resource quota support
 ```
 
-3. Deploy KfDef.
-```
+3. Deploy KfDef. _kfdef_ can point to a remote URL or to a local kfdef file. If you want to use the set of default kfdefs from Kubeflow. Follow the [Deploy with default kfdefs](#deploy-with-default-kfdefs) section below.
+```shell
 kubectl create -f <kfdef> -n ${KUBEFLOW_NAMESPACE}
 ```
-_kfdef_ above can point to a remote URL or to a local kfdef file. If you want to use the set of default kfdefs from Kubeflow. You will have to insert the `metadata.name` field before you can apply it to Kubernetes. Below are the commands for applying the KubeFlow 1.0 _kfdef_ using Operator. For e.g. for IBM Cloud, commands will be
+
+#### Deploy with default kfdefs
+To use the set of default kfdefs from Kubeflow, you will have to insert the `metadata.name` field before you can apply it to Kubernetes. Below are the commands for applying the Kubeflow 1.0 _kfdef_ using Operator. For e.g. for IBM Cloud, commands will be
 > If you are pointing the kfdef file on the local machine, set the `KFDEF` to the kfdef file path and skip the `curl` command.
 ```shell
 export KUBEFLOW_DEPLOYMENT_NAME=kubeflow
 export KFDEF_URL=https://raw.githubusercontent.com/kubeflow/manifests/master/kfdef/kfctl_ibm.yaml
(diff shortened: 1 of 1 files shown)


operator.md
```diff
@@ -16,18 +16,26 @@ kubectl create clusterrolebinding kubeflow-operator --clusterrole cluster-admin
 kubectl create -f deploy/operator.yaml -n ${OPERATOR_NAMESPACE}
 ```
 
-2. Deploy KfDef. You can optionally apply ResourceQuota if your Kubernetes version is 1.15+, which will allow only one _kfdef_ instance or one deployment of Kubeflow on this cluster, which follows the singleton model.
+2. Setup Kubeflow namespace. You can optionally apply ResourceQuota if your Kubernetes version is 1.15+, which will allow only one _kfdef_ instance or one deployment of Kubeflow on this cluster, which follows the singleton model.
```
animeshsingh:
```suggestion
2. Setup Kubeflow namespace. You can optionally apply ResourceQuota if your Kubernetes version is 1.15+, which will allow only one _kfdef_ instance or one deployment of Kubeflow on the cluster. This follows the singleton model, and is the current recommended and supported mode.
```

operator.md
```diff
@@ -16,18 +16,26 @@ kubectl create clusterrolebinding kubeflow-operator --clusterrole cluster-admin
...
 
-_<kfdef>_ above can point to a remote URL or to a local kfdef file. For e.g. for IBM Cloud, command will be
+3. Deploy KfDef.
+```
+kubectl create -f <kfdef> -n ${KUBEFLOW_NAMESPACE}
+```
+_kfdef_ above can point to a remote URL or to a local kfdef file. If you want to use the set of default kfdefs from Kubeflow. You will have to insert the `metadata.name` field before you can apply it to Kubernetes. Below are the commands for applying the KubeFlow 1.0 _kfdef_ using Operator. For e.g. for IBM Cloud, commands will be
 ```shell
-kubectl create -f https://raw.githubusercontent.com/kubeflow/manifests/master/kfdef/kfctl_ibm.yaml -n ${KUBEFLOW_NAMESPACE}
+export KUBEFLOW_DEPLOY_NAME=kubeflow
```
animeshsingh:
```suggestion
export KUBEFLOW_DEPLOYMENT_NAME=kubeflow
```

operator.md
```diff
@@ -16,18 +16,26 @@ kubectl create clusterrolebinding kubeflow-operator --clusterrole cluster-admin
...
+export KFDEF_URL=https://raw.githubusercontent.com/kubeflow/manifests/master/kfdef/kfctl_ibm.yaml
+export KFDEF=$(echo "${KFDEF_URL}" | rev | cut -d/ -f1 | rev)
+curl -L ${KFDEF_URL} > ${KFDEF}
+perl -pi -e $'s@metadata:@metadata:\\\n  name: '"${KUBEFLOW_DEPLOY_NAME}"'@' ${KFDEF}
```
animeshsingh:
```suggestion
perl -pi -e $'s@metadata:@metadata:\\\n  name: '"${KUBEFLOW_DEPLOYMENT_NAME}"'@' ${KFDEF}
```

animeshsingh: Do we need perl?

Tomcli: If we are going to use sed, then the commands will be different between osx and linux. If we want to have the same commands for both platform, usually people do it with awk or perl.

animeshsingh: ok - and will this work for the case where the kfdef is not through a URL, but locally on file system?

jlewi: If you really want to use "perl" or "sed" or "awk" I won't stop you; but I really don't like the idea of using an unstructured approach to process structured data.

I much prefer using a tool like yq
https://github.com/mikefarah/yq

Even though it means people have to install a separate binary I think it is a much better solution for modifying yaml files.

Tomcli: Thanks Jeremy, I updated the instructions to use yq. I kept the perl command in case someone is having trouble to install the yq binary.

PR #249, opened 2020-02-20
[CentralDashboard][Testing] End-to-end test linked as a post-submit test

avdaredevil:
## About
Added central-dashboard end-to-end test as a postsubmit check within endpoint ready test file. Note:
- Should this be a post-submit only test @jlewi?
- How can I run this file (seems like I cannot `pip install kubeflow.testing` like metadata, and I cannot find the relevant docs)? @jlewi or @richardsliu 

### Meta
/area testing
/area centraldashboard
/area front-end
/priority p0
/assign @avdaredevil
/cc @kwasi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/249)

### DIFF
--- a/py/kubeflow/kfctl/testing/pytests/endpoint_ready_test.py
+++ b/py/kubeflow/kfctl/testing/pytests/endpoint_ready_test.py
@@ -44,9 +44,14 @@ def test_endpoint_is_ready(record_xml_attribute, project, app_path, app_name, us
 # Todo(jlewi): Should this only run in postsubmit?
 @pytest.mark.skipif(os.getenv("JOB_TYPE") == "presubmit",
                     reason="test endpoint doesn't run in presubmits")
-def test_central_dash_is_running(record_xml_attribute, app_path):
+def test_central_dash_is_running(record_xml_attribute, app_path, kf_host, client_id, service_acct_email, service_acct_key_file):
   """Test that Kubeflow Central Dash was successfully deployed."""
   util.set_pytest_junit(record_xml_attribute, "test_central_dash_is_running")
+  os.environ["KF_HOST"] = kf_host       # The url to your cluster, usually: https://<cluster>.endpoints.<project>.cloud.goog/
+  os.environ["CLIENT_ID"] = client_id   # This is the CLIENT_ID token used to create the cluster (used in http://deploy.kubeflow.cloud/)
+  os.environ["SERVICE_ACCOUNT_EMAIL"] = service_acct_email  # The email address for the service account you created <name>@<project>.iam.gserviceaccount.com
+  os.environ["SERVICE_ACCOUNT_KEY"] = service_acct_key_file # Should be a JSON file downloaded from cloud console service accounts
+
   subprocess.check_call('cd "{}" && npm run test-e2e'.format(app_path), shell=True)
 
 if __name__ == "__main__":
--- a/py/kubeflow/kfctl/testing/pytests/endpoint_ready_test.py
+++ b/py/kubeflow/kfctl/testing/pytests/endpoint_ready_test.py
@@ -41,6 +41,14 @@ def test_endpoint_is_ready(record_xml_attribute, project, app_path, app_name, us
     if not gcp_util.iap_is_ready(url):
       raise Exception("IAP endpoint is not ready")
 
+# Todo(jlewi): Should this only run in postsubmit?
+@pytest.mark.skipif(os.getenv("JOB_TYPE") == "presubmit",
+                    reason="test endpoint doesn't run in presubmits")
+def test_central_dash_is_running(record_xml_attribute, app_path):
+  """Test that Kubeflow Central Dash was successfully deployed."""
+  util.set_pytest_junit(record_xml_attribute, "test_central_dash_is_running")
+  subprocess.check_call('cd "{}" && npm run test-e2e'.format(app_path), shell=True)
+
 if __name__ == "__main__":
   logging.basicConfig(level=logging.INFO,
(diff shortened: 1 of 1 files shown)


py/kubeflow/kfctl/testing/pytests/endpoint_ready_test.py
```diff
@@ -41,6 +41,19 @@ def test_endpoint_is_ready(record_xml_attribute, project, app_path, app_name, us
...
+                    reason="test endpoint doesn't run in presubmits")
+def test_central_dash_is_running(record_xml_attribute, app_path, kf_host, client_id, service_acct_email, service_acct_key_file):
+  """Test that Kubeflow Central Dash was successfully deployed."""
+  util.set_pytest_junit(record_xml_attribute, "test_central_dash_is_running")
+  os.environ["KF_HOST"] = kf_host       # The url to your cluster, usually: https://<cluster>.endpoints.<project>.cloud.goog/
+  os.environ["CLIENT_ID"] = client_id   # This is the CLIENT_ID token used to create the cluster (used in http://deploy.kubeflow.cloud/)
+  os.environ["SERVICE_ACCOUNT_EMAIL"] = service_acct_email  # The email address for the service account you created <name>@<project>.iam.gserviceaccount.com
+  os.environ["SERVICE_ACCOUNT_KEY"] = service_acct_key_file # Should be a JSON file downloaded from cloud console service accounts
+
+  subprocess.check_call('cd "{}" && npm run test-e2e'.format(app_path), shell=True)
```
jlewi: Per our discussion wrapping this in pytest is fine. I wonder if this is an unnecessary layer of direction and if its worth replacing right now or in a follow on PR.

PyTest is basically buying us the following functionality

1. It produces a junit XML file with the results which we need to surface results in prow
1. It is easy to mark the test as skippable?

It looks like the test is using the jasmine framework; does that framework provide this functionality?

I was originally thinking that by adding this test to endpoint_ready_test.py we could avoid the complexity of adding a step to the argo workflow. I suspect though we may not be able to avoid that. 

Since the test depends on IAP, we probably need to run this after the endpoint is ready test to make sure the IAP endpoint is available. Its possible there is a way to do this using pytest but otherwise we may need to modify the Argo graph.


/cc @kimwnasptd

avdaredevil:
@jlewi Can we merge this?

py/kubeflow/kfctl/testing/pytests/endpoint_ready_test.py
avdaredevil: Seems like it can: https://www.npmjs.com/package/jasmine-xml-reporter

jlewi:
@avdaredevil what about my previous question 

> Since the test depends on IAP, we probably need to run this after the endpoint is ready test to make sure the IAP endpoint is available. Its possible there is a way to do this using pytest but otherwise we may need to modify the Argo graph.

Is there a way to sequence the tests using pytest if not I think you will need to make it a separate step in the argo workflow.

avdaredevil:
I'd need the following information to fix this:
- KF_HOST
- CLIENT_ID
- SERVICE_ACCOUNT_EMAIL: # The email address for the service account you created <name>@<project>.iam.gserviceaccount.com
- SERVICE_ACCOUNT_KEY: # Should be a JSON file downloaded from cloud console service accounts

### Meeting Notes:
- Plan is to hardcode those values in the test (or still accept them as command line args
- Put the `gcp_util.iap_is_ready(url)` command in the dashboard test to make sure it doesn't run before cluster is ready

/cc @jlewi

py/kubeflow/kfctl/testing/pytests/endpoint_ready_test.py
```diff
@@ -41,6 +41,19 @@ def test_endpoint_is_ready(record_xml_attribute, project, app_path, app_name, us
...
+# Todo(jlewi): Should this only run in postsubmit?
+@pytest.mark.skipif(os.getenv("JOB_TYPE") == "presubmit",
+                    reason="test endpoint doesn't run in presubmits")
+def test_central_dash_is_running(record_xml_attribute, app_path, kf_host, client_id, service_acct_email, service_acct_key_file):
+  """Test that Kubeflow Central Dash was successfully deployed."""
+  util.set_pytest_junit(record_xml_attribute, "test_central_dash_is_running")
+  os.environ["KF_HOST"] = kf_host       # The url to your cluster, usually: https://<cluster>.endpoints.<project>.cloud.goog/
+  os.environ["CLIENT_ID"] = client_id   # This is the CLIENT_ID token used to create the cluster (used in http://deploy.kubeflow.cloud/)
+  os.environ["SERVICE_ACCOUNT_EMAIL"] = service_acct_email  # The email address for the service account you created <name>@<project>.iam.gserviceaccount.com
+  os.environ["SERVICE_ACCOUNT_KEY"] = service_acct_key_file # Should be a JSON file downloaded from cloud console service accounts
```
jlewi: Are you missing the call to:
gcp_util.iap_is_ready(url)?

jlewi:
To verify the test is working can you enable the test temporarily in the PR? (uncomment the code to skip it). Push it to GitHub to trigger a presubmit run. If the test passes include a link the PR thread as evidence the test is passing.

Then reenable the code to skip the test in presubmit and push it to the cluster.

PatrickXYS:
This is an useful test for functional E2E test

Since we haven't heard back from @avdaredevil , I'll go ahead to add it instead.

Jeffwan:
@PatrickXYS Feel free to take it over and you can use this as an inference. I will close this PR .

Issue #243, opened 2020-02-17
Segfault in the operator if the `manifests` repo is missing


vpavlin:
Just hit the following error [1]

The cause seems to be that I did not have a repo named `manifests` in the KFDef.

`kfctl` actually fails with

```
Error: failed to build kfApp from URI kfdef/kfdef_opendatahub_spark_only.1.0.yaml: couldn't generate KfApp:  (kubeflow.error): Code 500 with message: kfApp Generate failed for kustomize: kustomize generate failed Error: Repo manifests not listed in KfDef.Status; 
```

Is anyone aware why the name *manifests* is required? 

Can the operator at least not fail, but rather provide a reasonable error?

[1]
```
{"level":"info","ts":1581944513.3395808,"logger":"controller_kfdef","msg":"Reconciling KfDef","Request.Namespace":"kubeflow-spark","Request.Name":"spark-operator"}
{"level":"info","ts":1581944513.3396165,"logger":"controller_kfdef","msg":"Creating a new KubeFlow Deployment","Request.Namespace":"kubeflow-spark","Request.Name":"spark-operator","KubeFlow.Namespace":"kubeflow-spark"}
time="2020-02-17T13:01:53Z" level=info msg="Creating directory /tmp/.cache"
time="2020-02-17T13:01:53Z" level=info msg="Fetching https://github.com/vpavlin/odh-kf-manifests/archive/master.tar.gz to /tmp/.cache/odh-manifests"
time="2020-02-17T13:01:53Z" level=info msg="updating localPath to /tmp/.cache/odh-manifests/odh-kf-manifests-master"
time="2020-02-17T13:01:53Z" level=info msg="Fetch succeeded; LocalPath /tmp/.cache/odh-manifests/odh-kf-manifests-master"
time="2020-02-17T13:01:53Z" level=info msg="Repo manifests not listed in KfDef.Status; Resync'ing cache"
time="2020-02-17T13:01:53Z" level=info msg="/tmp/.cache/odh-manifests exists; not resyncing "
{"level":"info","ts":1581944513.4623666,"logger":"controller_kfdef","msg":"failed to build kfApp from URI ","Request.Namespace":"kubeflow-spark","Request.Name":"spark-operator","/tmp/config.yaml":null}
E0217 13:01:53.462589       1 runtime.go:73] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 1565 [running]:
[...]

PR #259, opened 2020-02-27
fix typo to avoid panic in operator controller

adrian555:
Toward #243 

The current kfctl does seems to expect a repo name `manifests` as the code shows https://github.com/kubeflow/kfctl/blob/9e8a07225b04d96eccf37dd0316e02a9a1e72f67/pkg/kfapp/kustomize/kustomize.go#L397 using the const https://github.com/kubeflow/kfctl/blob/c5c55f6d1b79e285f28fa433d6e4e2e739cefb63/pkg/apis/apps/group.go#L52.

This PR only fixed the specific code error in the operator side where the error was not caught due to a typo.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/259)


Tomcli:
@adrian555 I believe this just fix the error panic right? is it fixing the #243 with the repo name?

adrian555:
@Tomcli this only fixed the error panic in operator code as the title and description say. We can leave #243 open if we want to use it for supporting any repo name.

Tomcli:
ok thanks @adrian555 
/lgtm

PR #271, opened 2020-03-06
add ownerReference to resources when deployed through operator

### LINKED ISSUE #241: Kubeflow operator should not delete namespace it did not create
Quoting operator.md:

```
KUBEFLOW_NAMESPACE=kubeflow
kubectl create ns ${KUBEFLOW_NAMESPACE}
# kubectl create -f deploy/crds/kfdef_quota.yaml -n ${KUBEFLOW_NAMESPACE} # only deploy this if the k8s cluster is 1.15+ and has resource quota support
kubectl create -f <kfdef> -n ${KUBEFLOW_NAMESPACE}
```

When you do `oc delete -f <kfdef>` the operator deletes the namespace which it should not as it does not own it

adrian555:
Closes #241 

To avoid the deletion of namespace `kubeflow` that is not created as part of the kubeflow deployment through the operator, this PR adds the `ownerReference` to all resources in the manifest repo. The `KfDef` object is the owner of all resources in the `kustomize` directory. With this approach, when the `KfDef` kind of custom resource is deleted, unlike uninstalling kubeflow with `kfctl` CLI, the k8s garbage collector will kick in to delete all children resources and leave the namespace intact.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/271)


Tomcli:
/lgtm
Can we also open an issue to clarify that this delete won't delete user namespaces created by profile controller? Thanks.

adrian555:
Thanks @Tomcli those namespaces will not be deleted and the behavior is the same as kubeflow deleted through kfctl command line. So I added a note to the operator [readme](https://github.com/kubeflow/kfctl/blob/b025a6789748aa6bab4cbb5d14f586b29e7370bc/operator.md) instead.

PR #277, opened 2020-03-10
Create iap resources only when kfdef contains iap-ingress

kunmingg:
Currently kfctl create IAP resources by default and will throw error when `CLIENT_ID` not set.
User actually might customize kfdef and use private ingress.

This change makes kfctl creating iap resources only when kfdef contains iap-ingress

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/277)


Jeffwan:
Since GCP doesn't use kfctl now, I assume this PR is no longer needed. I will close the PR and feel free to reopen it if anything changed

PR #282, opened 2020-03-17
Find User Default Region and Cred Setting

PatrickXYS:
## Issue to solve
The issue is found in the kubeflow repo [here](https://github.com/kubeflow/kubeflow/issues/4854), the situation is that user use `kfctl apply kfctl_aws.yaml` can't find their plugin default settings.

## Changes we make
Rewrite `IsEksCluster` function to take default plugin settings without setting `AWS_REGION` manually.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/282)

### DIFF
--- a/pkg/kfapp/aws/aws.go
+++ b/pkg/kfapp/aws/aws.go
@@ -134,7 +134,10 @@ func GetPlatform(kfdef *kfconfig.KfConfig) (kftypes.Platform, error) {
 		},
 	}
 
-	session := session.Must(session.NewSession())
+	// set aws.sess with shared config file information, such as region
+	session := session.Must(session.NewSessionWithOptions(session.Options{
+		SharedConfigState: session.SharedConfigEnable,
+	}))
 
 	k8sClient, err := getK8sclient()
 	if err != nil {


pkg/kfapp/aws/eks.go
```diff
@@ -47,12 +49,26 @@ func (aws *Aws) getEksCluster(clusterName string) (*Cluster, error) {
...
 // IsEksCluster checks if an AWS cluster is EKS cluster.
 func (aws *Aws) IsEksCluster(clusterName string) (bool, error) {
+	session := session.Must(session.NewSession())
+
+	svc := sts.New(session)
+	stsinput := &sts.GetCallerIdentityInput{}
+	result, err := svc.GetCallerIdentity(stsinput)
+
+	if err != nil {
+		log.Infoln("AWS Credentials seems not correct %v", err.Error())
```
PatrickXYS: If we found that AWS Cred seems not correct with the error, should we then `return false, err`?  /cc @Jeffwan

Jeffwan: Hey, this is not the way we like to move forward. You'd better read region from config and create regional session and used by all AWS service client. 

Or use provider chain to find env, profile, ec2 metadata service. etc.

PatrickXYS: Make sense. Propose new version to improve this.

PR #288, opened 2020-03-23
kfctl should support using kustomize to compose a stack of applications

jlewi:
kfctl should support using kustomize to compose a stack of applications.

* Design doc: http://bit.ly/kf_kustomize_v3
* Issue: kubeflow/manifests#774

* Update kfctl to properly generate a kustomize application based on the
  Kubeflow application stack. The new application will use as a base
  the stack as provided in the KfDef.

Here's how this works. A KFDef would define a new application "kubeflow-apps"

```
apiVersion: kfdef.apps.kubeflow.org/v1
kind: KfDef
metadata:
  creationTimestamp: null
  name: stack_test
  namespace: kubeflow
spec:
  applications:
  - kustomizeConfig:
      repoRef:
        name: manifests
        path: stacks/gcp
    name: kubeflow-apps
```

* The "path" would point to a kustomize application that composes all of the packages to be deployed.
* Instead of copying this kustomize package to ${KFAPP}/kustomize, kfctl defines a new kustomize application that uses the stack as a base
  * The advantage of this approach is that to upgrade and pull in any changes the user just
     needs to refresh the .cache directory

* For this application we change the logic for setting application parameters; instead of modifying the parameters in the KFDef we generate a configmap patch for the application that sets the
  parameter.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/288)

### DIFF
--- a/pkg/kfconfig/types_test.go
+++ b/pkg/kfconfig/types_test.go
@@ -22,9 +22,13 @@ import (
 	"github.com/prometheus/common/log"
 	"io"
 	"io/ioutil"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"os"
 	"path"
+	"path/filepath"
 	"reflect"
+	"sigs.k8s.io/kustomize/v3/pkg/types"
 	"testing"
 )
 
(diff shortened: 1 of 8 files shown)


jlewi:
/assign @gabrielwen 
/assign @yanniszark 

The design was approved at this week's community meeting so this PR is ready to go.

pkg/kfconfig/types_test.go
```diff
@@ -435,6 +439,208 @@ func TestKfConfig_SetSecret(t *testing.T) {
...
+			},
+		},
+	}
+
+	for _, c := range cases {
+		addPatchStratgicMerge(c.Input, c.PatchName)
+		if !reflect.DeepEqual(c.Input, c.Expected) {
+			pGot, _ := Pformat(c.Input)
+			pWant, _ := Pformat(c.Expected)
+			t.Errorf("Error adding patch %v; got;\n%v\nwant;\n%v", c.PatchName, pGot, pWant)
```
gabrielwen: I think using `go-cmp` is better for readability.  it's pretty hard to see where went wrong with the literal printout.

jlewi: Done

jlewi:
@gabrielwen I made your changes could you please re-lgtm?

PR #291, opened 2020-03-24
GCP: remove dependecies from DM & IAM hack

kunmingg:
kfdef:
Make DM config & IAM hack config optional for platform == GCP:
* Currently during `kfdef build` & `kfdef apply`, kfdef assumes DM config is required when kfdef's platform == GCP, and would fill default value for DM config entry.  
With this PR we make DM config optional: kfdef only apply DM config when it is configured in kfdef. 
* Current `IAM hack` during `kfdef apply`: kfdef will clear GCP SA's roles and re-apply them to clean any existing leftover from previous deployment (and assume IAM template config exists) 
With this PR we run `IAM hack` only when IAM template config exists, since when KCC replaced DM, IAM roles leftover should be cleaned by KCC.

### DIFF
--- a/pkg/kfapp/gcp/gcp.go
+++ b/pkg/kfapp/gcp/gcp.go
@@ -697,28 +697,53 @@ func (gcp *Gcp) updateDM(resources kftypesv3.ResourceEnum) error {
 			Message: fmt.Sprintf("Error creating deploymentmanagerService: %v", err),
 		}
 	}
+	// DM entries are optional, defined in kfdef config. We only make DM changes when DM config was generated during kfctl build
 	if _, storageStatErr := os.Stat(path.Join(gcp.kfDef.Spec.AppDir, GCP_CONFIG, STORAGE_FILE)); !os.IsNotExist(storageStatErr) {
+		if storageStatErr != nil {
+			return &kfapis.KfError{
+				Code:    int(kfapis.INTERNAL_ERROR),
+				Message: fmt.Sprintf("Error creating deploymentmanager storage: %v", storageStatErr),
+			}
+		}
 		storageEntry, err := gcp.updateDeployment(deploymentmanagerService, gcp.kfDef.Name+"-storage", STORAGE_FILE)
 		if err != nil {
 			return kfapis.NewKfErrorWithMessage(err, fmt.Sprintf("could not update %v", STORAGE_FILE))
 		}
 		dmOperationEntries = append(dmOperationEntries, storageEntry)
 	}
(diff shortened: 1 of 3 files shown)


jlewi:
/assign @gabrielwen 

@kunmingg could you explain please what the DM and IAM hack are and how you are making them optional? What do we do instead?

kunmingg:
Description updated.

pkg/kfapp/gcp/gcp.go
```diff
@@ -727,62 +729,65 @@ func (gcp *Gcp) updateDM(resources kftypesv3.ResourceEnum) error {
...
-	appDir := gcp.kfDef.Spec.AppDir
-	gcpConfigDir := path.Join(appDir, GCP_CONFIG)
-	iamPolicy, iamPolicyErr := utils.ReadIamBindingsYAML(
-		filepath.Join(gcpConfigDir, "iam_bindings.yaml"))
-	if iamPolicyErr != nil {
-		return &kfapis.KfError{
-			Code: iamPolicyErr.(*kfapis.KfError).Code,
-			Message: fmt.Sprintf("Read IAM policy YAML error: %v",
-				iamPolicyErr.(*kfapis.KfError).Message),
+	if _, iamStatErr := os.Stat(path.Join(gcp.kfDef.Spec.AppDir, GCP_CONFIG, "iam_bindings.yaml")); iamStatErr == nil {
```
jlewi: Please add a comment explaining why we check if the file exists.

Should you check that iamStatErr is an error other than file doesn't exist and if it is return that error?

pkg/kfapp/gcp/gcp.go
```diff
@@ -704,11 +704,13 @@ func (gcp *Gcp) updateDM(resources kftypesv3.ResourceEnum) error {
 		}
 		dmOperationEntries = append(dmOperationEntries, storageEntry)
 	}
-	dmEntry, err := gcp.updateDeployment(deploymentmanagerService, gcp.kfDef.Name, CONFIG_FILE)
-	if err != nil {
-		return kfapis.NewKfErrorWithMessage(err, fmt.Sprintf("could not update %v", CONFIG_FILE))
+	if _, mainStatErr := os.Stat(path.Join(gcp.kfDef.Spec.AppDir, GCP_CONFIG, CONFIG_FILE)); mainStatErr == nil {
```
jlewi: nit: add a comment explaining the if condition.
Also it would probably be good to check  if the error is something other than file does not exist to return the error rather than treating it as the file not existing.

jlewi:
Thanks @kunmingg just a couple simple comments.

Issue #296, opened 2020-04-05
istio injection in Kubeflow namespace breaks KF 1.0 configs


jlewi:
#131 changed kfctl to set a label on the kubeflow namespace to turn on istio sidecar injection in the kubeflow namespace.

This change is not backwards compatible; i.e using an updated kfctl with the existing KFDef 1.0 manifests would cause Kubeflow 1.0 to stop working because kubeflow 1.0 is not comaptible with istio side car injection in the kubeflow namespace.

This is causing the tests on the kubeflow/manifests 1.0 branch to fail because they are testing with kfctl from master.

Proposed course of action:

1. Rollback #131 
1. Figure out a way to selectively enable istio side car injection only for Kubeflow configs >= 1.1

/assign @Bobgy

Bobgy:
Makes sense to me.

So we can guard the behavior of enabling istio injection in kfctl behind a config value in KfDef?
Who's the best person I should talk to? Should I try to contribute this one?

jlewi:
@Bobgy Yeah I think it would be great if you could pick this up. I think the difficulty will be deciding on a good pattern for implementing this.

Here are some of the options I can think of.

1. Adding an explicit field to KFDef for this
1. Adding logic based on the version/url of the manifests to decide whether its sufficiently high enough version to turn on istio
1. Applying KFDef's labels to the Kubeflow namespace
1. Adding a KFDef annotation
1. Create a new kustomize package that defines the namespace objects with associated labels

I don't think #1 is a good approach. In general, I want to remove logic and KFDef and kfctl and move kfctl in the direction of being dumb, syntactic sugar around kustomize. So I think this would be a step in the wrong direction.

#2 seems brittle.

For #3 I'm not sure how we would achieve consistent semantics.  In particular KF is potentially creating multiple namespaces (e.g. namespace for kubeflow components, istio namespace, knative, etc....).  I don't think its true that we would want to set istio side car injection in all those namespaces uniformly. So applying the labels to all the namespaces wouldn't work and I don't like the idea of hardcoding special treatment for certain namespaces.

#4 This is probably the obvious quick fix but it feels like a bit of a hack.

#5 I think I like this approach the best. Moving the namespaces we want created into YAML files as opposed to defining them in code seems consistent with the overall direction we want to move in. It also makes it really easy for users to customize. However, there might be some problems I'm not anticipating.

@yanniszark @johnugeorge @animeshsingh WDYT?

Bobgy:
[...]

PR #131, opened 2019-12-10
Enable istio injection in Kubeflow namespace

Bobgy:
/cc @jlewi 
Can you help me review?

Resolves  https://github.com/kubeflow/kubeflow/issues/3866
This PR is imitating https://github.com/kubeflow/kfctl/pull/123, which is similar

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfctl/131)

PR #297, opened 2020-04-06
Revert "Enable istio injection in Kubeflow namespace"

richardsliu:
Reverts kubeflow/kfctl#131

See https://github.com/kubeflow/kfctl/issues/296

### DIFF
--- a/pkg/utils/k8utils.go
+++ b/pkg/utils/k8utils.go
@@ -57,7 +57,6 @@ import (
 const (
 	YamlSeparator              = "(?m)^---[ \t]*$"
 	CertDir                    = "/opt/ca"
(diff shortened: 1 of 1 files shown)


jlewi:
Thanks @richardsliu 

@Bobgy @zhenghuiwang WDYT? Do you think reverting #131 is the right course of action?

zhenghuiwang:
> @Bobgy @zhenghuiwang WDYT? Do you think reverting #131 is the right course of action?

LGTM.

Bobgy:
I think postsubmit was run, because the PR was created from a branch not in forked repo.

kfp's postsubmit is configured to only run on master branch, you may do something similar

PR #300, opened 2020-04-10
Support KUBECONFIG environment in building k8s client

### LINKED ISSUE #298: aws/k8sClient doesn't respect KUBECONFIG env var
I had trouble installing Kubeflow, I tracked down the issue to the fact that it was trying to install on an old cluster I had torn down. This is because despite the fact that I was using the `KUBECONFIG` env var to point to the right kubeconfig, it wasn't respecting that. This led to weird results where I could `kubectl create` a namespace but `kfctl` would fail to create a namespace.

This is a regression from previous versions, as it has worked for me in the past.

The offending line is here, it should be updated to first check for KUBECONFIG: https://github.com/kubeflow/kfctl/blob/master/pkg/kfapp/aws/k8sClient.go#L18

Jeffwan:
Resolves #298 

Signed-off-by: Jiaxin Shan <seedjeffwan@gmail.com>

### DIFF
--- a/pkg/kfapp/aws/k8sClient.go
+++ b/pkg/kfapp/aws/k8sClient.go
@@ -14,8 +14,13 @@ import (
 
 // getK8sclient creates a Kubernetes client set
 func getK8sclient() (*clientset.Clientset, error) {
-	home := homeDir()
-	kubeconfig := filepath.Join(home, ".kube", "config")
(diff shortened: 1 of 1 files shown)


Jeffwan:
/cc @gabrielwen @kunmingg @jlewi 

This is a v1.0.2 release candidate

Issue #293, opened 2020-03-29
Improve the kfctl delete algorithm


yanniszark:
I'm opening this issue to explain what kfctl delete does right now, the problems users have bumped into and how we can make it better.

**What `kfctl delete` does right now:**
* Delete some predetermined global resources with a LabelSelector:
    * CRDs, ClusterRoleBindings, ClusterRoles
* Delete Kubeflow namespace

**Issues with current approach:**
* Relies on correct labels existing on Kubeflow resources.
* Asymmetric: resources are created with kustomize build, but are deleted with a LabelSelector and by deleting namespaces.
* Misses other namespaces of Kubeflow:
    * auth namespace for dex
    * knative namespace
    * istio namespace
    * cert-manager namespace
* Misses global resources:
    * cert-manager apiservice
* Many issues about buggy behaviour: https://github.com/kubeflow/kfctl/issues/56, https://github.com/kubeflow/kubeflow/issues/4807, https://github.com/kubeflow/kubeflow/issues/3767, https://github.com/kubeflow/kubeflow/issues/4486, https://github.com/kubeflow/kubeflow/issues/4496

**Proposed solution:**
We could patch the current code with more ways to find what's missing,
however I believe the best and most sustainable way to ensure delete works correctly is to follow a symmetric model.

Just like how in our dev workflows we do `kubectl apply -f` and `kubectl delete -f` which are symmetrical, `kfctl` should also follow the same symmetrical model.
I'm hoping that once we can agree on a design, development shouldn't be too hard.

cc @jlewi @vaskoj @DiscordianFish @gabrielwen

jlewi:
@yanniszark are you proposing that  `kfctl delete` is just sugar for `kustomize build -f | kubectl delete -`? 

Should we even have a `kfctl delete` command? At this point maybe we should keep it only for backwards compatibility. However, assuming `kfctl delete` is just a wrapper for `kustomize build -f | kubectal delete -` then maybe it would be better not to have `kfctl delete` as its an invitation for people to request/features different from `kubectl delete`.

The one reason I can think we might need it is to provide sugar for looping over all the kustomize packages and issuing a 

```
kustomize build -f | kubectl delete -
```

For each of them.

yanniszark:
> @yanniszark are you proposing that kfctl delete is just sugar for kustomize build -f | kubectl delete -?

Yeap! In the same way that apply is a wrapper for kustomize build -f | kubectl apply -f -
[...]

PR #309, opened 2020-04-14
kfctl delete refactor

yanniszark:
Fixes: #293 

Refactor the kfctl deletion algorithm as described in the issue.

### DIFF
--- a/pkg/kfapp/kustomize/kustomize.go
+++ b/pkg/kfapp/kustomize/kustomize.go
@@ -22,6 +22,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"io/ioutil"
+	errutil "k8s.io/apimachinery/pkg/util/errors"
 	"math/rand"
 	"os"
 	"path"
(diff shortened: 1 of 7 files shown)


yanniszark:
@adrian555 @Tomcli @kunmingg @jlewi this is ready for review.
I have tested it in my Kubeflow cluster multiple times.

pkg/kfapp/kustomize/kustomize.go
```diff
@@ -350,36 +346,84 @@ func (kustomize *kustomize) Delete(resources kftypesv3.ResourceEnum) error {
 				Message: msg,
 			}
 		}
+	}
```
adrian555, line 349: I wonder if line 339 to line 349 are still needed when line 361 below already handles the empty case.

pkg/kfapp/kustomize/kustomize.go
```diff
@@ -350,36 +346,84 @@ func (kustomize *kustomize) Delete(resources kftypesv3.ResourceEnum) error {
...
+				Message: fmt.Sprintf("error splitting yaml: %v", err),
+			}
+		}
+		for _, r := range resources {
+			err := utils.DeleteResource(r, kubeclient, 5*time.Minute)
+			if err != nil {
 				return &kfapisv3.KfError{
-					Code:    int(kfapisv3.INVALID_ARGUMENT),
-					Message: msg,
+					Code:    int(kfapisv3.INTERNAL_ERROR),
```
adrian555, line 417: Should we continue to delete the rest as a best effort or quit immediately?

yanniszark: That's a good question :thinking: 
I usually think it's good to fail early.
On the other hand, kubectl tries to delete everything regardless.
We may want to stay close to kubectl behavior as in the future people will probably use it.

pkg/kfapp/kustomize/kustomize.go
yanniszark: Nice catch, you're right!

yanniszark:
@adrian555 I made changes as per the review.
It's ready for another pass whenever possible :smile:

adrian555:
Looks good, thanks @yanniszark.

/lgtm

PR #321, opened 2020-04-27
Add Dockerfile.ubi and deploy by kustomize for operator

vpavlin:
This PR:

* Adds `Dockerfile.ubi` for UBI based operator image
* Make `Makefile` and `build-operator` more flexible to configure
  * `DOCKERFILE` for Dockerfile name
  * `IMAGE_BUILDER` to allow using `podman` to build the image
  * `OPERATOR_BINARY_NAME` - since `operator-sdk` uses the folder name for output controller binary, this variable parameterizes that (e.g. our fork is called `opendatahub-operator` and thus the binary name != `kfctl`)
* Add Kustomize for deployment manifests to simply run
  * `kustomize build deploy/` 
  * This allows simple extensibility and configurability for the deployment manifests (e.g. Open Data Hub will be able to rename some of the resources to `opendatahub-operator` without actually forking the manifests)

Should I also replace the `perl` line in Makefile with `kustomize set` ?

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,24 +1,25 @@
+ARG binary_name=kfctl
 FROM registry.access.redhat.com/ubi8/ubi-minimal:latest AS build
 
-ENV OPERATOR=/usr/local/bin/kfctl \
+ENV OPERATOR=/usr/local/bin/${binary_name} \
     USER_UID=1001 \
     USER_NAME=kfctl \
     HOMEDIR=/homedir
 
 # install operator binary
-COPY build/_output/bin/kfctl ${OPERATOR}
+COPY build/_output/bin/${binary_name} ${OPERATOR}
 
 COPY build/bin /usr/local/bin
 RUN  /usr/local/bin/user_setup
 RUN  /usr/local/bin/entrypoint
 
 FROM gcr.io/distroless/base-debian10
-ENV OPERATOR=/usr/local/bin/kfctl \
+ENV OPERATOR=/usr/local/bin/${binary_name} \
     USER_UID=1001 \
     USER_NAME=kfctl \
     HOMEDIR=/homedir
 
-COPY --from=build /usr/local/bin/kfctl /usr/local/bin/kfctl
+COPY --from=build /usr/local/bin/${binary_name} ${OPERATOR}
 COPY --from=build /etc/passwd /etc/passwd
 COPY --from=build ${HOMEDIR} ${HOME}
 
--- a/build/Dockerfile.ubi
+++ b/build/Dockerfile.ubi
@@ -1,0 +1,16 @@
+ARG binary_name=kfctl
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
+ENV OPERATOR=/usr/local/bin/${binary_name}\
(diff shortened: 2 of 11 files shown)


build/Dockerfile.ubi
```diff
@@ -0,0 +1,16 @@
+ARG binary_name=kfctl
```
adrian555, line 1: Since the change is more about OpenDataHub and with this `Dockerfile.ubi`, I wonder whether we should just leave the other `Dockerfile` as is. That is `Dockerfile.ubi` can be used to build both kubeflow and opendatahub operator images while `Dockerfile` is purely for kubeflow operator image.

deploy/base/cluster_role_binding.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
 subjects:
 - kind: ServiceAccount
   name: kubeflow-operator
-  namespace: ${OPERATOR_NAMESPACE} # Replace it with operator's namespace
+  namespace: $(namespace) # Replace it with operator's namespace
```
adrian555, line 9: doc https://github.com/kubeflow/kfctl/blob/master/operator.md needs update as well.

build/Dockerfile
```diff
@@ -1,24 +1,25 @@
+ARG binary_name=kfctl
 FROM registry.access.redhat.com/ubi8/ubi-minimal:latest AS build
 
-ENV OPERATOR=/usr/local/bin/kfctl \
+ENV OPERATOR=/usr/local/bin/${binary_name} \
     USER_UID=1001 \
     USER_NAME=kfctl \
```
adrian555, line 6: does USER_NAME need to change as well?

adrian555:
> Should I also replace the `perl` line in Makefile with `kustomize set` ?

I think we should. And that also means the dependence on `kustomize` needs to add to `Prerequisites` section in the [doc](https://github.com/kubeflow/kfctl/blob/master/operator.md).

Thanks @vpavlin

build/Dockerfile
vpavlin: The username does not really influence anything in the way how things are setup in the Dockerfile, IMO, so no need for that

build/Dockerfile.ubi
vpavlin: Ah, sure, we can. I just wanted to make sure things will behave consistently, but whatever is preferred

deploy/base/cluster_role_binding.yaml
vpavlin: Will do

vpavlin:
@adrian555 Thanks for the review, I updated the doc, please confirm if you want me to revert the changes in the `Dockefile`.

adrian555:
Thanks @vpavlin. Yes, please revert the changes in `Dockerfile`. Would like to have one set of things that works well as a kubeflow operator and then the other. As you can see now that the directory structure of `deploy` has changed and the support of kustomize, it may impact the auto generation of OLM CSV. Hence we have to be careful here. Thanks.

vpavlin:
Oh, shoot, I should have probably replied that I am reworking this:) I'll submit a follow up PR - got it working even with the CSV generation

Issue #242, opened 2020-02-17
Name the .cache and kustomize directories based on the KFDef name


vpavlin:
Currently Kubeflow operator will use `/tmp/.cache` and `/tmp/kustomize` to store the downloaded and generated manifests.

These directories should be ideally based on the KFDef resource name, since once the `/tmp/kustomize` exists, operator will skip the `kustomize.Generate` step:

```
time="2020-02-17T12:48:48Z" level=info msg="folder /tmp/kustomize exists, skip kustomize.Generate"
```

Imagine a scenario where:

1. A user submits KFDef
2. Realizes he needs to add another manifests repo to the list
3. Adds the manifests repository and runs `kubectl apply -f`
4. Operator fails with 
   ```
   time="2020-02-17T12:48:48Z" level=error msg="error evaluating kustomization manifest for component Error absolute path error in '/tmp/kustomize/component' : evalsymlink failure on '/tmp/kustomize/component'
   ```

   because it did not update the `/tmp/kustomize`

There are probably other scenarios - like someone adding a resource to KF manifests and the operator ignoring it, but I guess for development it is ok to just fall back to `kfctl` directly

vpavlin:
Hmm..now that I read my description, it really does not make much sense. Sorry for this.

IIRC I came across this issue when I submitted a KFDef names `X`, deleted it and submitted another - different - KFDef named `Y` where `Y = X + additional component`.

And I combined this problem with potentially another issue where we want to submit 2 KFdefs - in case, in the future, we have the ability to enable components on namespace scope.

So I would now say that there is this issue in the name - path to .cache and kustomize should be based on the resource name and then there is issue where the cache needs to be updated when the KFDef resource is updated by a user.

What do you think?

adrian555:
@vpavlin yes, I agree in a case when the kubeflow supports multiple instances within a namespace or a cluster (note that current operator is a cluster scoped). And based on just resource name may be solving the update partially. We can have a PR to append the KfDef name as I think it is a right thing.

I also agree the bigger problem is when the configuration of a certain component gets changed, but the KfDef name remains the same. Because the kustomization regeneration is bypassed, even though the KfDef resource itself is updated, the real kubeflow deployment remains the same.

Let the operator code always generate the kustomization directories and files may work for adding additional components. But what if a component is removed from the current KfDef resource? Still thinking how to handle these cases and whether the operator should have the similar behavior as command line deployment.

vpavlin:
My approach would be:

[...]

PR #324, opened 2020-05-01
tie kfapp directory with kfdef name and namespace

adrian555:
Towards #242 

This is the simpler fix for the above issue. It does following:

1. for a new kfdef instance, the kfapp directory is created as `/tmp/<namespace>/<name>`. The kfdef configuration file is copied to this directory. The cache and kustomize directories are also sub directories to this kfapp dir.

2. the manifest cache and kustomization files are downloaded and created by the same `kfApply` function like what `kfctl` command line does.

3. when a change in kfdef instance is caught, the `Reconcile` function will first remove this kfapp directory, then the same `kfApply` function runs to create/update according to the new kfdef configuration.

4. when a kubeflow resource is deleted, the same `kfApply` is run when the reconcilation kicks in. But this time, since the kfapp directory remains and the contents do not get changed, the deleted kubeflow resource will get recreated.

5. when the kfdef instance is deleted, the kfapp directory is destroyed as well.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/aws/aws-sdk-go v1.27.1
 	github.com/cenkalti/backoff v2.2.1+incompatible
 	github.com/chai2010/gettext-go v0.0.0-20170215093142-bf70f2a70fb1 // indirect
-	github.com/davecgh/go-spew v1.1.1
+	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/deckarep/golang-set v1.7.1
 	github.com/docker/docker v1.13.1 // indirect
 	github.com/docker/spdystream v0.0.0-20181023171402-6480d4af844c // indirect
@@ -36,7 +36,7 @@ require (
 	github.com/spf13/cobra v0.0.5
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.4.0
-	github.com/stretchr/testify v1.4.0
+	github.com/stretchr/testify v1.4.0 // indirect
 	github.com/tektoncd/pipeline v0.10.1
 	github.com/tidwall/gjson v1.4.0
 	github.com/tidwall/pretty v1.0.1 // indirect
@@ -49,7 +49,7 @@ require (
 	gopkg.in/airbrake/gobrake.v2 v2.0.9 // indirect
 	gopkg.in/gemnasium/logrus-airbrake-hook.v2 v2.1.2 // indirect
 	gopkg.in/yaml.v2 v2.2.8
-	gotest.tools v2.2.0+incompatible
+	gotest.tools v2.2.0+incompatible // indirect
 	k8s.io/api v0.17.0
 	k8s.io/apiextensions-apiserver v0.0.0
 	k8s.io/apimachinery v0.17.1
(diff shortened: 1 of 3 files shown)


adrian555:
cc @vpavlin @nakfour @Tomcli @animeshsingh

nakfour:
Thanks @adrian555, as I read this, it is deleting the kfapp even if only a change is detected, correct?

adrian555:
@nakfour correct with any change to kfdef.

animeshsingh:
cc @gabrielwen @jlewi

jlewi:
Thanks @animeshsingh for the heads up. I don't have much comment on the controller.
I think on GCP we are moving away from using a single monolithic operator to manage all of Kubeflow. So I probably won't be having much to say on any code specific to the operator.

On GCP we are moving in the direction of using GitOps and managing it as a loose collection of applications. Happy to discuss more at a community meeting if you are interested.

if you want to see where we are headed.
https://github.com/kubeflow/gcp-blueprints

If you look at the [makefile](https://github.com/kubeflow/gcp-blueprints/blob/master/kubeflow/Makefile)

You can see the pattern is

1. Hydrate manifests
1. Apply manifests

animeshsingh:
Thanks @jlewi - that's news! So does this imply on GCP you are also moving away from kfdef?

jlewi:
> So does this imply on GCP you are also moving away from kfdef?

It means that KFDef is becoming redundant with kustomization.yaml. i.e.  KFDef is just providing a list of kustomization packages to install. This can be done by listing kustomize packages in kustomization.yaml.

So I think whether we move away from it on GCP depends on whether we find a suitable purpose for having that extra layer of indirection.

vpavlin:
This looks good to me - I tested it and was able to deploy into 2 separate namespaces.

The operator reacted when I deleted one of the tracked configMaps and re-created it
The operator updated the deployment when I edited the kfdef
The deployment was correctly deleted based on ownerRef when I deleted the KFDef

Any case I missed and it should be tested? If not LGTM

adrian555:
Thanks @vpavlin for testing this out. I think the tests you have done covered the cases this PR fixes. I am writing some document according to what Animesh suggested above. Other than that, the code should be complete.

adrian555:
the `go-iap` presubmit seems to fail at the apply phase

```
INFO     root:util.py:72 Error: failed to apply:  (kubeflow.error): Code 500 with message: coordinator Apply failed for gcp:  (kubeflow.error): Code 500 with message: gcp apply could not
 update deployment manager: could not update storage-kubeflow.yaml; Insert deployment error: googleapi: Error 403: Quota exceeded for quota group 'deploymentMutators' and limit 'Write queries per day' of service 'deploymentmanager.googleapis.com' for consumer 'project_number:29647740582'., rateLimitExceeded
```

vpavlin:
So the problem is quota limit being hit by the CI? Can we do something about it? It is very unfortunate if this is the thing that blocks this PR from being merged.

adrian555:
@animeshsingh @vpavlin same error failing to create gcp client from the [log](http://testing-argo.kubeflow.org/workflows/kubeflow-test-infra/kubeflow-kfctl-presubmit-kfctl-go-iap-324-55090fd-4960-7d13?tab=workflow&nodeId=kubeflow-kfctl-presubmit-kfctl-go-iap-324-55090fd-4960-7d13-992711493). Could @richardsliu @jlewi please help take a look?

animeshsingh:
Thanks @jlewi for getting the quota increased to fix the CI issue
@adrian555 @vpavlin 

/lgtm
/approve

vpavlin:
Awesome, thank you very much

PR #326, opened 2020-05-06
Fix syncing cache behind HTTP proxy

AlexanderEkdahl:
This mimics the default `RoundTripper` and it uses `http.ProxyFromEnvironment` as directed by the $HTTP_PROXY and $NO_PROXY (or $http_proxy and $no_proxy) environment variables. This allows the `Transport` to work behind a proxy.

See https://golang.org/pkg/net/http/#Transport and https://golang.org/pkg/net/http/#RoundTripper for more details.

I assume registering custom protocol is due to internal requirements as it is not necessary when downloading files from GitHub.

### DIFF
--- a/pkg/kfconfig/types.go
+++ b/pkg/kfconfig/types.go
@@ -516,7 +516,9 @@ func (c *KfConfig) SyncCache() error {
 				return errors.WithStack(err)
 			}
 		} else {
-			t := &http.Transport{}
+			t := &http.Transport{
+				Proxy: http.ProxyFromEnvironment,
+			}
 			t.RegisterProtocol("file", http.NewFileTransport(http.Dir("/")))
 			t.RegisterProtocol("", http.NewFileTransport(http.Dir("/")))
 			hclient := &http.Client{Transport: t}


AlexanderEkdahl:
@googlebot I signed it!

pkg/kfconfig/types.go
```diff
@@ -516,7 +516,9 @@ func (c *KfConfig) SyncCache() error {
 				return errors.WithStack(err)
 			}
 		} else {
-			t := &http.Transport{}
+			t := &http.Transport{
+				Proxy: http.ProxyFromEnvironment,
```
kkasravi, line 520: is there a way to provide a test for this?

AlexanderEkdahl: This can certainly be tested and the behavior is tested in detail here: https://github.com/golang/net/blob/master/http/httpproxy/proxy_test.go. However, I don't think it is valuable to test this in kfctl as it introduces undue maintenance burden on the kfctl developers. 

Using `http.ProxyFromEnvironment` is the default behavior for all HTTP request in Go([link](https://golang.org/src/net/http/transport.go#L42)) and this PR simply ensures the same behavior as the transport happens to be overriden.

AlexanderEkdahl:
@Jeffwan I can't make sense of the apparent test failure. From what I can tell all of the tests passed?

PatrickXYS:
@AlexanderEkdahl 

Can you rebase and run tests?

Jeffwan:
/lgtm
/approve

Thanks for the contribution

shalberd:
How come this is not present in the master branch?

Issue #57, opened 2019-10-16
kfctl is overly reliant on expensive E2E tests; slowing down development


jlewi:
kfctl is overly reliant on expensive E2E tests.

I'm seeing presubmits taking 50 minutes to run. Furthermore as these E2E tests become more comprehensive they inherently become more flaky.

I think we need to rethink our test strategy for kfctl to ensure velocity remains high.

* For example, how can we develop a component test for kfctl that doesn't end up being a complete E2E test for kubeflow?
 
   * e.g. Does it really make sense when testing changes to kfctl to test that all the Kubeflow applications are deployed and healthy?

jlewi:
One idea would be to

* Selectively trigger tests specific to plugins when the plugin changed
* Use purposely designed kustomize applications and KFDef files to test the bulk of kfctl functionality

The bulk of kfctl functionality is just about  processing a KFDef to deploy a bunch of kustomize applications.

We shouldn't really need to use actual Kubeflow manifests or create clusters to test that.

We could create simple kustomize manifests that do really cheap and reliable operations; e.g. create a Job that just runs "echo Hello world". 

That could easily run in the test cluster (where prow jobs run); as opposed to creating a new cluster.

Plugin authors (e.g. authors of gcp.go, aws.go, existing.go) could still trigger more expensive comprehensive tests selectively when those plugins are modified.

jlewi:
@swiftdiaries this is the issue tracking trying to come up with a better way of testing kfctl.

jlewi:
Should we punt this to post 1.0? I don't think anyone is working on this  and I'm not sure we will have cycles to fix it before we want to release 1.0.

jlewi:
Also I think the direction we want to move is kfctl should mostly be focused on helping users build manifests and not applying/deleting them. apply/delete(#293) should mostly be synonymous with kubectl apply/delete.

So kfctl testing should mostly focus on generating the config files and validating them but not necessarily applying them and verifying they produce a working deployment. 

So similar to #1016 kfctl tests should mostly be running `kfctl build` and validating the produced configs match the expected configs.  This would eliminate a huge source of presubmit flakiness.

PR #329, opened 2020-05-13
Add --dump flag to build

discordianfish:
When specified, apply will dump all manifests to stdout instead of
applying them.

This is useful to validate changes to kfdef and kubeflow upgrades.


discordianfish:
Tests fail because:
```
Insert deployment error: googleapi: Error 403: Quota exceeded for quota group 'deploymentMutato
rs' and limit 'Write queries per day' of service 'deploymentmanager.googleapis.com' for consumer 'project_number:29647740582'., rateLimitExceeded
```

discordianfish:
Looks like this isn't working since it breaks the order in which the resources get applied. Looking into that now.

discordianfish:
Ok this should do it!

jlewi:
Thanks @discordianfish 

Is there anyway to achieve this functionality using off the shelf functionality rather than continuing to build out a Kubeflow specific CLI?

IIUC the problem is that we have a directories of independent kustomize packages and we'd like to run `kustomize build` in all those directories.

If I had to stack rank my preferred solutions it would be

1. kustomize
1. kpt
1. kfctl

I think this can be done via a [kpt function](https://googlecontainertools.github.io/kpt/reference/fn/source/) which is now upstreamed into kustomize at least on master.

Here's an example.
https://github.com/kubernetes-sigs/kustomize/blob/266af31bcb7b471e5135ccd8d35b1d24d09158a9/functions/examples/injection-tshirt-sizes/image/main.go

Could we adapt that to

1. Find all kustomization.yaml files
1. Apply the kustomization.yaml and emit the resulting files

discordianfish:
@jlewi But I need this to work with my existing 1.0.1 setup, so I can compare the output to whatever next version we will deploy.

In general yeah, I could do without kfctl in general and only use kubeflow stock tools. But again, whatever they will do, I like to compare the current kfctl generated manifests to it.

I mean, unless there is a better way to validate that an update is safe, given that updating to the next kubeflow release won't be a few line change in my kfdef but more involved.

That being said, I now have this functionality so it doesn't really matter to me whether to merge this.

jlewi:
I think this is valuable and don't want to block a good enough solution.

Should we be using "build" and not "apply" though to generate hydrated manifests? Maybe something like

```
kfctl build -dump
```

jlewi:
@discordianfish at @yanniszark thoughts?

I have a strong preference for this to be a subcommand of build not apply. This seems more consistent with the semantics of kustomize and a shift left philosophy. Under a shift left philosophy build would produce hydrated manifests (which is effectively what this command) is doing and this would then be piped to the apply stage which could be kfctl, kubectl, or other tools like ArgoCD?

vpavlin:
My 2c... `--dry-run` for `apply` does not make much sense semantically since dry running apply would mean to me that you are trying to verify that all the produced resources are applicable in the cluster.

If the goal is to generate all the resources to be able to compare multiple revisions, then the suggested `--dump` makes perfect sense to me.

yanniszark:
Have we discussed this before? I'm having a deja-vu moment :sweat_smile: 
But yes, I agree this makes more sense with build.
`--dry-run` implies some sort of validation (and is what kubectl does)
I see this has been implemented as a separate function `Dump` so it should be a matter of changing a couple of files under `cmd` (unless I have missed something).

discordianfish:
Yeah no disagreement from my side, just haven't gotten to it. Looking into that now.

jlewi:
Looks like metadata test failed; probably unrelated to this PR. 
kubeflow/kfctl#57 is tracking getting rid of the overreliance on flaky E2E tests.

/test all

cmd/kfctl/cmd/build.go
```diff
@@ -45,27 +45,30 @@ var buildCmd = &cobra.Command{
...
+				return fmt.Errorf("couldn't generate KfApp: %v", err)
 			}
-			return nil
-		} else {
+		default:
 			return fmt.Errorf("Unsupported object kind: %v", kind)
 		}
 
-		return err
+		if applyCfg.GetBool(string(kftypes.DUMP)) == true {
```
jlewi, line 68: Should this be `buildCfg`  if we are binding build?

cmd/kfctl/cmd/build.go
```diff
@@ -94,4 +97,13 @@ func init() {
 		log.Errorf("Couldn't set flag --%v: %v", string(kftypes.VERBOSE), bindErr)
 		return
 	}
+
+	// dump flag
+	applyCmd.Flags().BoolP(string(kftypes.DUMP), "d", false,
```
jlewi, line 102: Should this be buildCmf and buildCfg if the command is build not apply?

discordianfish:
Oops, yeah copy&paste mess. Fixed now. Also fixed the commit message.

jlewi:
/lgmt
/approve

Thanks; unfortunately looks like there is merge conflict.

jlewi:
THanks!
/lgtm
/approve

PR #335, opened 2020-05-21
A kustomize function to remove namespace from specific custom resources.

jlewi:
* See kubeflow/gcp-blueprints#27 We use a kustomize function
  to remove namespace from specific kubeflow resources which
  are instances of cluster scoped resources which should not have
  a namespace set.

  * This is needed in order to create configurations compatible with
     ACM.
    
* This code should be considered experimental.
    
   * It is an initial attempt at using kustomize functions which is still pretty
     new.


jlewi:
/assign @yanniszark 
/assign @animeshsingh 

@yanniszark @animeshsingh could you PTAL? In particular do you have any objections with committing this code which is still "experimental" into the kfctl repository?

animeshsingh:
I am good. 
/lgtm

jlewi:
Thanks @animeshsingh 

/approve

PR #338, opened 2020-05-27
Update OWNERS

pdmack:
Add some Red Hat approvers and reviewers to support repo activity


jlewi:
Per
https://www.kubeflow.org/docs/about/contributing/#maintaining-owners-files

Can we grow the nested OWNERs files as opposed to root level?

vpavlin:
@jlewi Do you mean to add a specific OWNERs file for the operator ?

pdmack:
I'm not sure where that nesting would go.

jlewi:
So I might suggest

* Create an owner file within each directory one level down from root
   * e.g. kfctl/pkg, kfctl/hack, etc...
* We can use an OWNER alias https://www.kubeflow.org/docs/about/contributing/#owners_aliases
   to avoid repetition if we have directories that should be owned by all the same folks (e.g. tests)
* In cases where it does make sense to have more narrowly defined owners files we should do that.
* So the operator director is an obvious choice

* Some of the subdirectories in https://github.com/kubeflow/kfctl/tree/master/pkg/kfapp would also make sense

vpavlin:
Kfctl does not feel like such a big project that it would need owners per directory, but maybe I am missing something here

jlewi:
> Kfctl does not feel like such a big project that it would need owners per directory

Per: https://www.kubeflow.org/docs/about/contributing/#maintaining-owners-files 
More OWNERs files is a good thing

* More OWNERs files makes it easier to autoassign reviews to the appropriate folks.

* As kfctl grows there are starting to emerge specialized pieces where it makes sense to have narrowly scoped OWNERs files to ensure the right people approve the code e.g. 

  * operator - I'm not particularly familiar with the operator so it makes sense to not include me as a primary OWNER of the operator code
  * platform specific code - platform OWNERs should be approving this code
  * kustomize-fns - This is currently isolated from the rest of kfctl

* A side effect of the above is that we probably have to create OWNERs files in each subdirectory
  * This is an unfortunate quirk but I think its preferable to having a growing list of root OWNERs
  * IMO we should aim for root OWNERS to have 2-3 people to
    i) act as backup approvers in the event people go AWOL
   ii) Be willing and able to properly assign PRs to the right subfolks

OWNERS
```diff
@@ -2,10 +2,13 @@
...
   - animeshsingh
+  - crobby
   - gabrielwen
   - kkasravi
   - kunmingg
   - jlewi
+  - vpavlin
   - yanniszark
 reviewers:
+  - pdmack
```
jlewi, line 13: alphabetize

jlewi:
Ping?

@vpavlin and @crobby is one of you willing to drive the kfctl release #352?

crobby:
> Ping?
> 
> @vpavlin and @crobby is one of you willing to drive the kfctl release #352?

Sure, I'll give it a go.

animeshsingh:
Thanks @crobby 

@adrian555 will help out as well, so you both can co-lead the release process for kfctl/manifests, as they are linked together... 

unless we have someone identified for manifests @jlewi ?

animeshsingh:
as for now, I am ok with PR, and we can come back in future to look for owner aliases and nested files.

/lgtm

animeshsingh:
@crobby @vpavlin you both need to be in members list for kubeflow, so please do a PR here
https://github.com/kubeflow/internal-acls/blob/master/github-orgs/kubeflow/org.yaml

vpavlin:
@animeshsingh I see both of us as the kubeflow members in the yaml - do we need to add ourselves to a specific subgroup?

OWNERS
```diff
@@ -1,8 +1,15 @@
 approvers:
   - adrian555
   - animeshsingh
+  - crobby
+  - gabrielwen
+  - kkasravi
+  - kunmingg
```
jlewi, line 7: Looks like this is outdated due to #353

PR #342, opened 2020-06-01
kfctl mirror should recourse over dependencies

### LINKED ISSUE #340: kfctl mirror images should not hardcode the kustomize directory
Currently the mirror code assumes there is a directory named "kustomize" and then walks that directory to identify images to be mirror'd.
https://github.com/kubeflow/kfctl/blob/e96de9d7fd614f5c78b9e19609bfb2849e8297f1/pkg/mirror/mirror_image.go#L260

This doesn't work with GCP blueprints because the actual files area layed out differently. There's really no reason for the mirror command to assume the directory is named "kustomize". Instead we should add a command line flag to specify the directory and default to the current working directory.

jlewi:
kfctl mirror should recourse over dependencies

* The current implementation of kfctl mirror was not looking
  at dependent packages that were out of tree. This change
  causes the mirror command to look at the resources directive
  inside the kustomization.yaml and follow those links.

  * This is necessary to support blueprints and some of the refactoring
    of the kustomize manifests intended to better support Day 2 operations.

* Fix #340 kfctl mirror shouldn't hard code the directory to search
  for images but instead allow it to be provided via a flag.

### DIFF
--- a/cmd/kfctl/cmd/mirror_build.go
+++ b/cmd/kfctl/cmd/mirror_build.go
@@ -15,12 +15,16 @@ import (
 )
 
 var outputFileName string
+var directory string
 var gcb bool
 
 func init() {
 	replicateBuildCmd.Flags().StringVarP(&outputFileName, "output", "o", "",
 		`Name of the output pipeline file
 		kfctl alpha mirror build -o <name>`)
+	replicateBuildCmd.Flags().StringVarP(&directory, "directory", "d", "kustomize",
+		`The directory to search for kustomization files listing images to mirror
+		kfctl alpha mirror build -d <directory>`)
 	replicateBuildCmd.Flags().BoolVar(&gcb, "gcb", false, `Generate cloud build config`)
 	// verbose output
 	replicateBuildCmd.Flags().BoolP(string(kftypes.VERBOSE), "V", false,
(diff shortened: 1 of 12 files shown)


jlewi:
Actually it looks like we can parameterize the functions:
https://googlecontainertools.github.io/kpt/reference/fn/run/

So we shouldn't actually need to add annotations to all the objects we can just add an appropriate parameter to the function.

jlewi:
/hold

I'm going to split this up into two PRs

1. PR to add flags to the mirror command
1. PR to add transforms to replace image prefixes.

jlewi:
@yanniszark this is ready for review. I scoped down the PR as mentioned in my earlier comment.

yanniszark:
@jlewi thanks a lot for taking the time to improve `kfctl`.
Unfortunately, I don't have the cycles right now to review this PR and I don't want to be the one blocking it.
It would be very helpful if another community member could take up the review of this PR.

jlewi:
/assign @Tomcli 

@Tomcli Could you review this please?

/unassign @yanniszark

adrian555:
@jlewi sorry for the delay, reviewed above. Thanks.

jlewi:
Thanks @adrian555 

I ran go fmt.
I believe there are issues open to do this as part of presubmits.

PR #345, opened 2020-06-02
kustomize fn to transform images

jlewi:
* This PR defines a kustomize/kpt function to transform images based on their prefix.

  * This is an alternative to using kfctl mirror overwrite

  * This works as follows.

    i) you define a function config like the following

       ```
        apiVersion: v1alpha1 # Define a transform to change all the image prefixes to use images from a different registry
     kind: ImagePrefix
     metadata:
       name: use-mirror-images-gcr
       annotations:
         config.kubernetes.io/function: |
           container:
             image: gcr.io/kubeflow-images-public/kpt-fns:v1.0-rc.3-58-g616f986-dirty
     spec:
       imageMappings:
       - src: quay.io/jetstack
         dest: gcr.io/gcp-private-dev/jetstack # {"type":"string","x-kustomize":{"setBy":"kpt","partialSetters":[{"name":"gcloud.core.project","value":"gcp-private-dev"}]}}
       - src: gcr.io/kubeflow-images-public
         dest: gcr.io/gcp-private-dev # {"type":"string","x-kustomize":{"setBy":"kpt","partialSetters":[{"name":"gcloud.core.project","value":"gcp-private-dev"}]}}
       ```

   ii) You use kpt to appy the function e.g.

       ```
       kpt fn ${DIR}
       ```

       * Typically you would apply this to the hydrated manfiests
       * The transform will look for any objects with `image-prefix.kubeflow.org` annotation
         and then map the image prefixes specified in that annotation

* The kpt function has a number of advantages over the kfctl functionality

  * In particular the transform configuration (i.e. the annotation) doesn't need to change
    when a user upgrades and the image tags change. Since the transform is applied after
    hydration we would end up applying the transform to any docker images

  * `kpt fns` are more scalable then baking all the functionality into a single go binary

    * It also looks like this functionality will be upstreamed into kustomize

* kubeflow/manifests#1033 proposed a similar transformer for images.

  * This is different only in we are using kpt/kustomize functions rather than kustomize transformations
  * kpt functions are more flexible because its YAML in YAML out; so you could just as easily


jlewi:
The e2e test failures look like flakiness for which I have #346 
/test all

jlewi:
Unittests failed.

```
# github.com/kubeflow/kfctl/v3/cmd/kfctl/cmd
cmd/kfctl/cmd/mirror_build.go:82:42: too many arguments in call to mirror.GenerateMirroringPipeline
        have (string, "github.com/kubeflow/kfctl/v3/pkg/apis/apps/imagemirror/v1alpha1".ReplicationSpec, string, bool)
        want ("github.com/kubeflow/kfctl/v3/pkg/apis/apps/imagemirror/v1alpha1".ReplicationSpec, string, bool)
# github.com/kubeflow/kfctl/v3/cmd/kfctl/cmd
vet: cmd/kfctl/cmd/mirror_build.go:82:43: cannot use directory (variable of type string) as v1alpha1.ReplicationSpec value in argument to mirror.GenerateMirroringPipeline
Makefile:71: recipe for target 'vet' failed
make: *** [vet] Error 2
```

jlewi:
The build failure was due to files I failed to revert and is now fixed.

adrian555:
@jlewi, see review above, thanks.

jlewi:
> input could be argoproj/argoui:v2.3.0 for pipelines deployment.

```
src: argoproj.io
dest: gcr.io/myproject/argoproj.io
```

I didn't add a Make file because I didn't think it was necessary but can add one if you like.

adrian555:
@jlewi it is fine without a Makefile. Thanks.

PR #346, opened 2020-06-02
Do not run KF E2E tests as blocking presubmits.

jlewi:
* Per kubeflow/kfctl#57 this removes the expensive E2E tests as blocking
presubmits. Instead of attempting to deploy Kubeflow and verify its working
we should aim to have good unittests that have a high degree of confidence.

* E2E testing of actual deployments should happen downstream and be the
  provenance of platform owners.

* The E2E tests can still run as periodic and postsubmit tests.

### DIFF
--- a/prow_config.yaml
+++ b/prow_config.yaml
@@ -21,10 +21,12 @@ workflows:
       - pkg/*
       - testing/*
       - py/*
+  #****************************************************************************************************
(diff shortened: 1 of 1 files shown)


adrian555:
Thanks @jlewi looks good.

/lgtm

PR #347, opened 2020-06-03
Rename target `build` to `build-kfctl`

MrXinWang:
To fix CI error (https://github.com/kubeflow/manifests/pull/1210#issuecomment-637914317) caused by target name change.

Signed-off-by: Henry Wang <Henry.Wang@arm.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@
 # limitations under the License.
 #
 GCLOUD_PROJECT ?= kubeflow-images-public
-GOLANG_VERSION ?= 1.12.7
+GOLANG_VERSION ?= 1.13.7
 GOPATH ?= $(HOME)/go
 GOOS ?= $(shell go env GOOS)
 GOARCH ?= $(shell go env GOARCH)
@@ -58,7 +58,7 @@ JUNIT_FILE ?= /tmp/report.xml
 
 %.md:
 
-all: build
+all: build-kfctl
 
 auth:
 	gcloud auth configure-docker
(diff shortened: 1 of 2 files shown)


MrXinWang:
@Bobgy PR created

Makefile
```diff
@@ -58,7 +58,7 @@ JUNIT_FILE ?= /tmp/report.xml
 
 %.md:
 
-all: build
+all: build-kfctl
```
Bobgy, line 61: Instead of doing so many changes, I think you can just
```
build-kfctl: build
```
Did I miss anything?

MrXinWang: Well the reason why I removed this `build` target is because it does the samething exactly as `build-kfctl` and it seems redundant. If we bring that back it will be the same....

Bobgy: Understood, I was more thinking of this as a hotfix to unblock manifests presubmit test, so I prefered the change to be smaller (one line fix), but this also doesn't seem risky. If you think `build-kfctl` is a good name for this target for the long term, I don't have any objections.
/lgtm

MrXinWang: @Bobgy I checked the kubeflow/testing scripts and it seems that I have to submit another PR to change the binary name too, as currently the master branch makes the name of the binaries as `kfctl-<OS>-<ARCH>` instead of a simple `kfctl`...

Bobgy: Shall we revert the original PR instead of forward fixing then?
You can decide

MrXinWang: Yes maybe revert is a better idea before we identified how this will break everything...I will try to submit another PR later to add multiarch support again then.

MrXinWang:
closing in faver of #348

PR #350, opened 2020-06-04
fix error when config file is from local relative directory

adrian555:
When the `repos.uri` is a local directory following command fails, eg.

`kfctl build -V -f kfctl_istio.yaml`

```could not sync cache. Error: Rel: can't make .cache/manifests relative to /xxx/go/src/github.com/kubeflow/manifests```

This PR provides the fix for the problem.

### CODE REVIEW

**Comment by adrian555** [comment]
@jlewi @kkasravi could one of you please review? Thanks.

File: pkg/kfconfig/types.go [hunk 1]
```diff
@@ -504,12 +504,23 @@ func (c *KfConfig) SyncCache() error {
 		// Manifests are local dir
 		if fi, err := os.Stat(r.URI); err == nil && fi.Mode().IsDir() {
 			// check whether the cache directory is a sub directory of manifests
-			if relDir, err := filepath.Rel(r.URI, cacheDir); err != nil {
+			absCacheDir, err := filepath.Abs(cacheDir)
```
jlewi, line 507: Could we unittest this somehow?

adrian555: Thanks. Added one test case.

adrian555:
@jlewi could you please approve again? I had to merge the master to remove the presubmit test that was removed recently. Thanks.

Issue #304, opened 2020-04-11
Upgrades in 1.1 should follow kustomize off the shelf workflow


jlewi:
Filing this issue to track simplifying the upgrade process in Kubeflow 1.1.

Here's the current instructions for how Kubeflow upgrades are done.
https://www.kubeflow.org/docs/upgrading/upgrade/

This differs from the standard off the shelf workflow for kustomize applications
https://github.com/kubernetes-sigs/kustomize/blob/master/docs/workflows.md#off-the-shelf-configuration

In particular, we introduce a KFUpgrade resource which defines pointers to the old and new KFDef.
https://www.kubeflow.org/docs/upgrading/upgrade/#upgrade-instructions

kfctl then does a lot of a magic in order to try to reapply any user defined kustomizations ontop of the new configs.

With the new kustomize patterns (http://bit.ly/kf_kustomize_v3) we should be able to simplify this and I think eliminate the need for kfctl. Instead users should be able to just

1. Update .cache to point to a new version of the kubeflow/manifests directory
1. Run kustomize build to regenerate the package.

This is because the new pattern with stacks is that kfctl generates a new kustomize package using Kubeflow defined packages in .cache as the base. So a user can regenerate .cache without losing any of their kustomizations.

There are a couple of issues that we run into when applying the updated manifests

1. Pruning - how do we cleanup resources from earlier versions of Kubeflow that are no longer in the latest Kubeflow resource
1. Updating immutable fields - Certain fields are immutable and will cause errors when `apply` is called.

Rather than rely on kfctl logic to solve these problems we should follow a shift left pattern. Our expectation should be that we rely on existing tools (e.g. kubectl, kpt, etc...) to apply the manifests and handle these problems.

kpt for example supports [pruning](https://googlecontainertools.github.io/kpt/reference/live/)

/cc @richardsliu @yanniszark @kunmingg

jlewi:
As noted in #4873; kustomize commonLabels should only be used for immutable labels
https://kubectl.docs.kubernetes.io/pages/app_management/labels_and_annotations.html

As these are used in selectors which are immutable.

Right now our applications are using version in the version and instance label which is used in selector and set via commonLabels.
https://github.com/kubeflow/manifests/blob/abc6898ba535515e88846e7cc97faa208ffdacb9/jupyter/jupyter-web-app/overlays/application/kustomization.yaml#L11

We need to fix this so that labels will be immutable across version updates.

It looks like `kubectl` has alpha support in prune.

```
[...]

PR #351, opened 2020-06-05
Add the ability to create CloudEndpoint resources using kfctl.

jlewi:
* This is GCP specific code that allows CloudEndpoints to be created using
  the CloudEndpoint controller. A Cloud endpoint is a KRM style resource
  so we can kust have `kfctl apply -f {path}` invoke the appropriate
  logic.

* For GCP this addresses kubeflow/gcp-blueprints#36; specifically when
  deploying private GKE the CloudEndpoints controller won't be able
  to contact the servicemanagement API. This provides a work around
  by running it locally.

* This pattern seems extensible; i.e. other platforms could link in
  code to handle CR's specific to their platforms. This could basically
  be an alternative to plugins.

* I added a context flag to control the kubecontext that apply applies to.
  Unfortunately, it doesn't look like there is an easy way to use
  that in the context of applying KFDef. It looks like the current logic
  assumes the cluster will be added to the KFDef metadata and then
  look up that cluster in .kubeconfig.

  * Modifying that logic to support the context flag seemed riskier
    then simply adding a comment to the flag.

* Added some warnings that KFUpgrade is deprecated since per #304
  we want to follow the off shelf workflow.


adrian555:
@jlewi review is done above. 

I do have one general question. From the description above I am not clear whether there will be one or two commands to deploy Kubeflow. 

If as you said that this can be an alternative to the plugins, then I would suspect it will take two commands (or, two runs of `kfctl` command) to deploy. The new command option creates the CloudEndpoints with the `kubeContext`. But then if you want to deploy the Kubeflow, you will need to run the command with a KfDef configuration. And this time, you have to make sure that the KUBECONFIG is correctly pointed to the same cluster. 

If I get it right, my concern would be whether this model is sustainable in the long term. I probably need to read more on the CloudEndPoint controller. But not quite sure about the reason to have this `ep.Process` wrapped in the `kfctl`. I would rather expect that one command to complete two functionalities (eg. `kfctl apply -f <config> -p <ep-config> --kubeconfig <cluster-config>`).

Hope this does not add more confusion. :)

jlewi:
@adrian555 Thanks for the thoughtful response. 

So the first problem is as follows:

We as GCP want/need to ship additional CLI functionality to support our customers. The question is whether we should bake this into the existing kfctl or ship a new GCP specific binary.

Our preference would be to ship it as part of kfctl and not require users to install a separate CLI.

I have tried to do it in a way that is extensible to other performs and Kubernetes native. In particular, the syntax

```
kfctl apply -f  ${FILE}
```

means we aren't adding a new flag sub command. So in this regard I think its not a one off hack for `GCP`. i.e. if other platforms wanted to follow this approach they could do so and we wouldn't have to worry that we would end up with 1000 flags and commands that nobody would understand.

There is a risk of binary bloat and possibly conflicting version dependencies but arguably that's not a new problem since we already allow platform specific logic to be baked into kfctl.

> I do have one general question. From the description above I am not clear whether there will be one or two commands to deploy Kubeflow.

On GCP there will be multiple commands to deploy kubeflow. Initially we are using [Makefile](https://github.com/kubeflow/gcp-blueprints/blob/master/kubeflow/Makefile) to glue these commands together. My hope is that we will find a more cloud native solution to glue it together. That could be a KF specific solution (e.g. some iteration on the current approach) but I'm hoping a more generic, cloud native solution will emerge. [skaffold](https://skaffold.dev/docs/pipeline-stages/) for example has a very pipeline centric view but it currently wouldn't work for our use cases.

> If I get it right, my concern would be whether this model is sustainable in the long term

This is any area where multiple oppionions are possible and it may not be a place where one size fits all.

My personal oppinion is that treating it as a composable build pipeline makes it easier for folks to customize to meet their particular needs.

Fundamentally our process is two step

```
build | apply
```

Increasingly we see the need for folks to introduce additional steps in the pipeline. So with this pr the process is

```
build | apply | create DNS entry
```

adrian555:
Thanks @jlewi for the detailed info!

I totally agree on the pipelining aspect and we already see there are multiple sub commands/functions such as `alpha mirror`, `set-image-name` etc. So I am all in with the intention to bake things related with Kubeflow with `kfctl`.

One thing I may point out is the example of `build | apply | create DNS entry` may not be the case for `GCP` however since it is now using the kustomize to build and apply. :)

So my problem is more about whether the ep.Process should be part of the `apply` command. And I agree it depends on individual's opinion on how to interpret it. To me, running `apply` command means to apply the KfDef configuration of a set of Kubeflow applications. The result or output of it is a deployment of Kubeflow. Then if we compare the ep with KfDef plugins, we either treat that as an add-on (which can be a new `kfctl addon` command) or extra-config (which can be `kfctl apply -p <ep-config>`.

jlewi:
> Then if we compare the ep with KfDef plugins, we either treat that as an add-on (which can be a new kfctl addon command) or extra-config (which can be kfctl apply -p <ep-config>.

Both of those options diverge from kubectl semantics which is what we've modeled kfctl on.

The kubernetes approach to bulk operations is to specify a single directory or file with multiple resources.

We don't need to add new subcommands or flags precisely because the KRM (i.e. metadata) allows the binary to figure out what resource to handle it appropriately. 

So if we really wanted to we could support the having a single file with both resources separated by "---".

adrian555:
/lgtm

Thanks @jlewi .

PR #354, opened 2020-06-16
Update endpoint_ready_test for blueprints

jlewi:
Support multi-package namespaces for python.

Fix code to work with IAP using workload identity.

* The existing code to get an ID token didn't seem to work with workload
  identity and didn't match the latest code on the IAP. The latest
  code appears to use a helper function to get the id token

Related to: kubeflow/gcp-blueprints#51 Create a tekton test for blueprints
to verify the endpoint is ready.

### DIFF
--- a/py/kubeflow/kfctl/testing/util/gcp_util.py
+++ b/py/kubeflow/kfctl/testing/util/gcp_util.py
@@ -1,82 +1,66 @@
-import argparse
-import base64
 import datetime
 import logging
 import os
-import errno
-import shutil
-import subprocess
-import tempfile
-import threading
-from functools import partial
-from multiprocessing import Process
 from time import sleep
 from google.auth.transport.requests import Request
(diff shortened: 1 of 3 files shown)


py/kubeflow/kfctl/testing/util/gcp_util.py
```diff
@@ -101,14 +147,15 @@ def iap_is_ready(url, wait_min=15):
...
-          "GET",
-          url,
-          headers={
-              "Authorization":
-              "Bearer {}".format(google_open_id_connect_token)
-          },
-          verify=False)
+      #resp = requests.request(
+          #"GET",
+          #url,
```
Bobgy, line 152: nit: shall we get rid of the commented lines?

jlewi:
Thanks @Bobgy ; fixed

/approve

PR #356, opened 2020-06-17
handle generator with envs

adrian555:
This will fix the problem exposed by this PR https://github.com/kubeflow/manifests/pull/1204.

Add the handling of `envs` as the environment sources so `kfctl` can append the overlays' path to the `params.env` and `secrets.env` etc.

An example manifest is https://github.com/kubeflow/manifests/blob/master/metadata/overlays/db/kustomization.yaml

```
configMapGenerator:
- name: metadata-db-parameters
  envs:
  - params.env
secretGenerator:
- name: metadata-db-secrets
  envs:
  - secrets.env
```


adrian555:
@jlewi PTAL. I think this needs to get in asap.

jlewi:
Thanks.

/lgtm
/approve

PR #327, opened 2020-05-11
pretty print log messages IAM policy

Anamika28:
Existing pretty print function is utilised in places, where logs need to be printed in json format.

PR #362, opened 2020-06-23
Pretty print log messages IAM policy(from #327)

Jeffwan:
Code comes from #327, didn't hear back from original user for a while. I'd like to have this merged and released in next version.


Jeffwan:
@adrian555 

Can you help leave a /lgtm and /approve?  Seems you have the permission and part of release team. Thanks!

PR #372, opened 2020-07-08
Fix kf_is_ready_test

jlewi:
* istio_egressgateway is no longer deployed so need to remove it from
  list of deployments we check for; it doesn't seem to be required
  to enable outbound internet access.

* Split up the kf_is_ready_test into more focused tests

* Related to kubeflow/gcp-blueprints#73


jlewi:
Looks like unittests failed because we are out of disk space on our NFS mount.

jlewi:
No specific error; looks like a timeout downloading go modules
/test all

PR #375, opened 2020-07-10
Fix tests for GCP blueprints 1.1

jlewi:
* For workload identity binding tests the permissions were all wrong
* We no longer download service account keys
* cloud-endpoints-controller should be in the kubeflow namespace not
  istio-system namespace

Related to kubeflow/gcp-blueprints#52
Related to kubeflow/gcp-blueprints#73


Bobgy:
/lgtm

looks great

PR #379, opened 2020-07-22
Improve AWS kfctl

PatrickXYS:
### What change we make:

1. Deleted legacy code in `aws.go`

2. Since `github.com/kubernetes-sigs/application` deleted tag `v0.8.0` (https://github.com/kubernetes-sigs/application/tags), upgrade to `v0.8.1`.

### What test we run:
Run `make build` and it succeeded, then use new-built kfctl to deploy kubeflow and passed.

### DIFF
--- a/pkg/kfapp/aws/aws.go
+++ b/pkg/kfapp/aws/aws.go
@@ -454,10 +454,6 @@ func (aws *Aws) Generate(resources kftypes.ResourceEnum) error {
 		return errors.WithStack(err)
 	}
 
-	if err := aws.kfDef.SetApplicationParameter("istio-ingress", "namespace", IstioNamespace); err != nil {
-		return errors.WithStack(err)
-	}
-
 	if pluginSpec.Auth != nil && pluginSpec.Auth.BasicAuth != nil && pluginSpec.Auth.BasicAuth.Password != "" {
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -24,7 +24,7 @@ require (
 	github.com/hashicorp/go-version v1.2.0
 	github.com/imdario/mergo v0.3.8
 	github.com/jlewi/cloud-endpoints-controller v0.0.0-20200604211613-aff0aaad5602
-	github.com/kubernetes-sigs/application v0.8.0
+	github.com/kubernetes-sigs/application v0.8.1
```
Jeffwan, line 27: @PatrickXYS any reason to upgrade application version?  I didn't see following code is related to this?

PatrickXYS: Removed

pkg/kfapp/aws/aws.go
```diff
@@ -454,10 +454,6 @@ func (aws *Aws) Generate(resources kftypes.ResourceEnum) error {
 		return errors.WithStack(err)
 	}
 
-	if err := aws.kfDef.SetApplicationParameter("istio-ingress", "namespace", IstioNamespace); err != nil {
```
Jeffwan, line 457: good. This is fixed in istio-system. we don't need this anymore

PR #380, opened 2020-07-24
Use istio-parameter in manifests

PatrickXYS:
For now, `istio` component has not moved to V3 yet. Thus, it still use `istio-parameter` rather than `istio-config`.

On kfctl side, we have to use `istio-parameter` now in V3 changes.

Will roll back until `istio` moved to V3.

### DIFF
--- a/pkg/kfapp/aws/aws.go
+++ b/pkg/kfapp/aws/aws.go
@@ -467,10 +467,6 @@ func (aws *Aws) Generate(resources kftypes.ResourceEnum) error {
 		}
 	} else {
 		if pluginSpec.Auth.Cognito != nil {
-			if err := aws.kfDef.SetApplicationParameter("istio", "clusterRbacConfig", "ON"); err != nil {
-				return errors.WithStack(err)
-			}
-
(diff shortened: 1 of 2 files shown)


pkg/kfapp/aws/aws.go
```diff
@@ -467,9 +467,10 @@ func (aws *Aws) Generate(resources kftypes.ResourceEnum) error {
 		}
 	} else {
 		if pluginSpec.Auth.Cognito != nil {
-			if err := aws.kfDef.SetApplicationParameter("istio", "clusterRbacConfig", "ON"); err != nil {
-				return errors.WithStack(err)
-			}
+			// Will roll back after istio move to v3 done
```
Jeffwan, line 470: Do we need to roll back later?

PatrickXYS: We don't have to, it depends on if we want to set `clusterRbacConfig=ON` in kfctl logic or manifests configmap.

PatrickXYS:
@jlewi @adrian555 Can you take a look? We should be good to go.

adrian555:
@PatrickXYS if this is good from AWS platform side, please have someone LGTM. I will approve after that. Thanks.

Jeffwan:
@adrian555 Can you approve the change? Thanks

PR #388, opened 2020-07-31
Turn the remove-namespace kustomize function into a proper function.

jlewi:
* Use a client side CRD so we can easily add more types
  for which we need to remove the namespace.

* Package it in the same docker image as our image-prefix kpt function.

Fix kubeflow/gcp-blueprints#27


jlewi:
Not ready for review but creating a PR so I don't forget about it.

jlewi:
/assign @Bobgy 

This is ready for review.

PR #390, opened 2020-08-03
Create a kustomize function to set the gateway on virtual services.     

jlewi:
* The gateway on our virtual services is misconfigured as described in
  kubeflow/gcp-blueprints#22 and kubeflow/manifests#1169

  * We should be using the istio gateway in the istio namespace and not
    the gateway in the kubeflow namespace

* To begin to fix that this pr adds a kustomize function to transform
  all the gateways in the virtual services. This makes it possible for
  configurations to start transitioning to the gateway in the istio namespace

 * We need this for GCP blueprints with ACM (kubeflow/gcp-blueprints#22)
   because we don't want to duplicate the gateway in that case.

* Clean up some comments in the existing kustomize functions.

### DIFF
--- a/kustomize-fns/image-prefix/function.go
+++ b/kustomize-fns/image-prefix/function.go
@@ -4,7 +4,6 @@ package image_prefix
 import (
 	"fmt"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
-	"os"
 	"sigs.k8s.io/kustomize/kyaml/kio"
 	"strings"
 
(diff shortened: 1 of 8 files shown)


jlewi:
#388 should be merged first.

kustomize-fns/image-prefix/function.go
```diff
@@ -99,8 +98,6 @@ func (f *ImagePrefixFunction) Filter(inputs []*yaml.RNode) ([]*yaml.RNode, error
 // replaceImage looks for an annotation changing docker image prefixes and if it is present
 // applies it to all the images.
 func (f *ImagePrefixFunction) replaceImage(r *yaml.RNode) error {
-	m , _  := r.GetMeta()
```
vpavlin, line 102: How is this change related to this PR?

kustomize-fns/virtual-service/function.go
```diff
@@ -0,0 +1,112 @@
...
+
+var _ kio.Filter = &VirtualServiceFunction{}
+
+// Filter returns a new VirtualServiceFunction
+func Filter() kio.Filter {
+	return &VirtualServiceFunction{}
+}
+
+// VirtualServiceFunction implements the ImagePrefix Function
+type VirtualServiceFunction struct {
```
vpavlin, line 26: Would it make sense to use Istio client-go here instead of defining the structs? https://github.com/istio/client-g

jlewi: The structure isn't a virtual service or other defined ISTIO struct so I don't think we can use client go.

kustomize-fns/image-prefix/function.go
jlewi: Miscellaneous cleanup. Not directly related.

Issue #412, opened 2020-09-21
Dumping generate invalid yaml


jtherin:
Using kfctl 1.1.0 and build from master

> kfctl build -d -f https://raw.githubusercontent.com/kubeflow/manifests/master/kfdef/kfctl_k8s_istio.v1.1.0.yaml

generate an invalid yaml with empty line return instead of `---` between applications

Example:
```apiVersion: v1
kind: Namespace
metadata:
  labels:
    control-plane: kubeflow
    istio-injection: enabled
  name: kubeflow

apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  creationTimestamp: null
  labels:
    app.kubernetes.io/component: kubeflow
    app.kubernetes.io/name: kubeflow
  name: applications.app.k8s.io
spec:
  group: app.k8s.io
  names:
    kind: Application
    plural: applications
  scope: Namespaced
  validation:
    openAPIV3Schema:
      properties:
        apiVersion:
          type: string
        kind:
...
```

(missing `---` between Namespace resource and CustomResourceDefinition resource)

jbottum:
Hi @jtherin did you get this resolved ?  which platform (GCP, AWS, On-Prem) are you deploying on to ?

jtherin:
I'm deploying on-prem. I pushed https://github.com/kubeflow/kfctl/pull/413 to fix it.

PR #413, opened 2020-09-21
fix: dump generate invalid yaml

jtherin:
Fixes https://github.com/kubeflow/kfctl/issues/412

### DIFF
--- a/pkg/kfapp/kustomize/kustomize.go
+++ b/pkg/kfapp/kustomize/kustomize.go
@@ -217,6 +217,7 @@ func (kustomize *kustomize) Dump(resources kftypesv3.ResourceEnum) error {
 			return err
 		}
 		fmt.Println(string(data))
+		fmt.Println("---")
 	}
 	return nil
 }


pkg/kfapp/kustomize/kustomize.go
```diff
@@ -216,7 +216,8 @@ func (kustomize *kustomize) Dump(resources kftypesv3.ResourceEnum) error {
 		if err != nil {
 			return err
 		}
-		fmt.Println(string(data))
+		fmt.Print(string(data))
```
PatrickXYS, line 219: Add `---` is okay, but I do want to keep printing original data with new line.

PatrickXYS (commented):
@jtherin Thanks for contributing! 

I left a few comments

pkg/kfapp/kustomize/kustomize.go
jtherin: Okay, I've updated the PR :).

PR #419, opened 2020-10-26
[IBM] limit reconcile requests to specific kfdef update events and a couple of operator fixes

adrian555:
This PR handles these issues:

1. when there is an update event caught for the KfDef CR, check whether they are one of the three changes to the KfDef that the operator is supposed to handle. Others will be ignored. This is done to prevent any false update to the KfDef that could cause the operator to redeploy the Kubeflow.

2. fix the logic to check whether a namespace already exists and whether the kfdef label should be applied or not.

3. change the master branch's operator image tag to `:latest` so it will not impact the current release.


adrian555:
cc @Tomcli @pvaneck @moficodes

pkg/kfapp/kustomize/kustomize.go
```diff
@@ -1422,14 +1422,19 @@ func GenerateYamlWithOperatorAnnotation(resMap resmap.ResMap, instance *unstruct
...
 			}
-			_, err = corev1client.Namespaces().Get(m.GetName(), metav1.GetOptions{})
+			ns, err := corev1client.Namespaces().Get(m.GetName(), metav1.GetOptions{})
 			if err == nil {
 				log.Infof("Namespace %v already exists.", m.GetName())
 
-				_, found := anns[kfdefAnn]
-				if !found || anns[kfdefAnn] != kfdefCr {
-					// if the namespace is not created by this operator, should not append the annotation
+				nsAnns := ns.GetAnnotations()
```
pvaneck, line 1429: Just so I understand, the `anns := m.GetAnnotations()` on line 1409 did not get the namespace annotations properly, so you needed this?

adrian555: correct. That line got the annotation from the kustomize file not the current instance.

moficodes (approved):

PR #431, opened 2020-11-06
Disable broken AWS ALB ingress status check

PatrickXYS:
For now, the AWS ALB ingress status check is broken, which is described in this issue https://github.com/kubeflow/testing/issues/779,

Disable the status check until above issue fixed

/cc @Jeffwan @yanniszark


Jeffwan:
Latest stable version is 12.0. I would suggest to upgrade k8s python client and build new image. 
Removing the logic means we don't even check entrypoint which make test case not that helpful.

PatrickXYS:
Send another PR to fix the issue 

https://github.com/kubeflow/kfctl/pull/432

PatrickXYS:
/close 

close the PR in favor of  https://github.com/kubeflow/kfctl/pull/432

PR #444, opened 2020-11-13
Add configuration for eks_cluster_version

PatrickXYS:
To establish Kubeflow - Kubernetes compatibility matrix, add configuration for `eks_cluster_version`,

meanwhile set up presubmit test for 1.15 - 1.18 EKS cluster. 

/hold 

Hold until test succeed, technically we won't need to change kfctl's `prow_config.yaml` but in manifests `prow_config.yaml`

### DIFF
--- a/py/kubeflow/kfctl/testing/ci/kfctl_e2e_workflow.py
+++ b/py/kubeflow/kfctl/testing/ci/kfctl_e2e_workflow.py
@@ -74,10 +74,10 @@ class Builder(object):
                use_basic_auth=False,
                build_and_apply=False,
                test_target_name=None,
+               eks_cluster_version="1.17",
                extra_repos="",
                **kwargs):
     """Initialize a builder.
-
     Args:
       name: Name for the workflow.
       namespace: Namespace for the workflow.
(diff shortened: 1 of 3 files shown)


PatrickXYS:
image

It's working properly

PatrickXYS:
I think it's reasonable to update `go.mod` package `github.com/kubernetes-sigs/application` 

from `0.8.0` to `0.8.1`, since the package deleted `0.8.0` release previously https://github.com/kubernetes-sigs/application/releases

/cc @Jeffwan

go.mod
```diff
@@ -93,7 +93,7 @@ replace (
 	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.0.0-20190620090156-2138f2c9de18
 	k8s.io/metrics => k8s.io/metrics v0.0.0-20190620085625-3b22d835f165
 	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.0.0-20190620085408-1aef9010884e
-	sigs.k8s.io/application => sigs.k8s.io/application v0.0.0-20190404151855-67ae7f915d4e
+	github.com/kubernetes-sigs/application => sigs.k8s.io/application v0.0.0-20190404151855-67ae7f915d4e
```
Jeffwan, line 96: in this case. it's still replaced by old version. 

you can either help new version, especially package imports path, or just remove go.mod changes and submit AWS related changes in this PR. We can leave application package there until there's a problem

Jeffwan (changes requested):
I would suggest remove application module update, or remove field in `replace` and test new version.

go.mod
PatrickXYS: Reverted application module update

Jeffwan (approved):

Issue #422, opened 2020-10-29
[Operator]Add operator manifests for 1.2 release


adrian555:
Tasks:
[ ] add 1.2.0 operator csv https://github.com/kubeflow/kfctl/tree/master/deploy/olm-catalog/kubeflow, make sure the proper kfdef CR is used in `alm-examples`
[ ] update current version https://github.com/kubeflow/kfctl/blob/master/deploy/olm-catalog/kubeflow/kubeflow.package.yaml
[ ] build operator image for 1.2.0 and update https://github.com/kubeflow/kfctl/blob/master/deploy/operator.yaml to use that image
[ ] update version # in https://github.com/kubeflow/kfctl/blob/master/cmd/manager/main.go#L35
[ ] update https://github.com/kubeflow/kfctl/blob/master/operator.md to include the instructions for 1.2.0
[ ] fix file name https://github.com/kubeflow/kfctl/blob/master/deploy/olm-catalog/kubeflow/1.1.0/kubeflow.v1.0.0.clusterserviceversion.yaml

Jeffwan:
@adrian555 Any progress there? is this a blocking issue for kubeflow 1.2 release?

evan-hataishi:
@Jeffwan I'll submit a PR for all this by the end of the week :)

Jeffwan:
@evan-hataishi Do you need this to be in 1.2? We plan to cut the release this week.

evan-hataishi:
@Tomcli What do you think?

Tomcli:
@evan-hataishi can you update the version in https://github.com/kubeflow/kfctl/blob/master/cmd/manager/main.go#L35?

Everything else can be done after 1.2 release since olm yaml doesn't need to change the kfctl source code.

PR #448, opened 2020-11-16
Fix csv version

evan-hataishi:
Fixes the file name as described in one of the tasks in https://github.com/kubeflow/kfctl/issues/422.

CSV generated with operator-sdk 0.16
Tested on OCP 4.5 with this [kfdef](https://github.com/moficodes/manifests/blob/master/kfdef/kfctl_openaihub_openshift_tekton.yaml)


Tomcli:
@evan-hataishi so no code changes for 1.1 release except the filename?

evan-hataishi:
> @evan-hataishi so no code changes for 1.1 release except the filename?

@Tomcli Nope, just a fix for the filename.

Tomcli:
@evan-hataishi how about for 1.2? I think we will have the official 1.2 release this week.

evan-hataishi:
@Tomcli It looks like we're keeping track of all the versions, so I'll work on that next and put it in a new PR

Tomcli:
I see, it looks like the olm Kubeflow 1.1 release was using the old name. This probably fix that bug.
/lgtm

Jeffwan:
Seems this is OLM only. 

/approve

PR #451, opened 2020-11-18
Fixed the default output to point to `1.2` instead of `1.0`

naveensrinivasan:
The default command line output was still pointing to `1.0`. 

I updated it to point to `1.2` to the latest version. 

I couldn't find an update for `1.2` for this `	gcpConfig      = "https://raw.githubusercontent.com/kubeflow/manifests/v1.0-branch/kfdef/kfctl_gcp_iap.v1.0.0.yaml"`

### DIFF
--- a/cmd/kfctl/cmd/apply.go
+++ b/cmd/kfctl/cmd/apply.go
@@ -35,10 +35,10 @@ var kubeContext = ""
 
 // KFDef example configs to be printed out from apply --help
 const (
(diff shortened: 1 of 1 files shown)


Jeffwan:
@naveensrinivasan Thanks. This will be used in the hints if commands encounter some errors. Can you help cherry-pick this change?

naveensrinivasan:
> @naveensrinivasan Thanks. This will be used in the hints if commands encounter some errors. Can you help cherry-pick this change to v1.2-branch?

The hints were wrong and that is the reason I had to do a PR. I will cherry-pick this change to v1.2-branch.

PR #474, opened 2021-01-13
[Test] Replace unittest ksonnet code with python based

### LINKED ISSUE #430: Get rid of ksonnet in the tests
ksonnet was deprecated over a year ago the tests should no longer be using ksonnet to define tests.

Related to #774

PatrickXYS:
Resolves #430

As part of removing ksonnet work, I replaced existing unittest which is written by ksonnet with python unit test.

/hold

Hold until the tests passes


PatrickXYS:
Failed for `make: *** No rule to make target 'go-unittest-junit'.  Stop.`

Fix for restarting

PatrickXYS:
Confirmed tests passed, clean up left code.

PatrickXYS:
@theofpa Can you help review the code?

Makefile
```diff
@@ -303,6 +303,12 @@ test-junit: build-kfctl
 	echo Running tests ... junit_file=$(JUNIT_FILE)
 	go test ./... -v 2>&1 | go-junit-report > $(JUNIT_FILE) --set-exit-code
 
+
```
theofpa, line 306: Clean-up target `test-junit` above, previously used by ksonnet.

theofpa (changes requested):

Makefile
PatrickXYS: Addressed in latest commit

theofpa (approved):

PR #477, opened 2021-01-21
Disable kfctl e2e test because current manifests repo refactoring

PatrickXYS:
Since there will be large refactoring work in kubeflow/manifests repo, results in master branch's manifest fail to work, thus, disable kfctl e2e test to avoid PR blockage.

More details can be found https://github.com/kubeflow/manifests/pull/1719#issuecomment-763947044

/cc @yanniszark

### DIFF
--- a/prow_config.yaml
+++ b/prow_config.yaml
@@ -1,5 +1,5 @@
-# This file configures the workflows to trigger in our Prow jobs.
-# see kubeflow/testing/py/run_e2e_workflow.py
+## This file configures the workflows to trigger in our Prow jobs.
+## see kubeflow/testing/py/run_e2e_workflow.py
 python_paths:
   # Need to place kubeflow/testing in front of kubeflow/testing so that the package can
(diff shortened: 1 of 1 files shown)


yanniszark:
@PatrickXYS thanks for opening the PR!
Should we merge now or wait for a blockage to present and then merge?

PatrickXYS:
@yanniszark I think we can merge it as of now, because we've already predict it will happen for sure.

yanniszark:
@PatrickXYS sounds good.
It would be great if you could comment out the presubmit (instead of deleting) and add a comment with the reason why it is disabled.

/lgtm
/hold

Adding a hold in case you want to add the change above. If you think it's ok, then you can choose to remove the hold.

PatrickXYS:
@yanniszark Sounds good, added comment to explain the reason. 

/unhold

PatrickXYS:
@yanniszark Seems like you need to lgtm again

PR #490, opened 2021-03-17
Add ownerReferences to resources

### LINKED ISSUE #460: Ownerreferences no longer added to resources deployed through operator
Hello, it seems this feature was initially added [here](https://github.com/kubeflow/kfctl/pull/271) but was later removed [here]( https://github.com/kubeflow/kfctl/commit/20f20a739cb449db2c362347210ee0ead088f2e0#diff-0cd2fe44eca7e038192e34261390737f73795dc333b6b8cdf9bc008e8d6f57bb) in exchange for annotations. I also can't seem to find why this was changed. We found ownerreferences useful because we use ArgoCD to deploy our kfdefs and ArgoCD uses owner-references to draw out relationships (and possibly other logic) between the resources being deployed by the operator. Now Argocd only sees the kfdef and none of the components' resources being deployed. e.g.

image
vs: 
image

VaishnaviHire:
Closes #460 
This commit is responsible for following changes:

- Adds owner references to all the resources in the manifest repo.
- It also removes the `kfctl.kubeflow.io/kfdef-instance: <kfdef-name>.<kfdef-namespace>`
annotation, since its no longer required. The owner references field now indicates that
resources are owned by KfDef instance.

Ref PR: https://github.com/kubeflow/kfctl/pull/271

### DIFF
--- a/operator.md
+++ b/operator.md
@@ -107,14 +107,13 @@ The Kubeflow operator also support multiple _KfDef_ instances deployment. It wat
 
 The operator responds to following events:
 
-* When a _KfDef_ instance is created or updated, the operator's _reconciler_ will be notified of the event and invoke the `Apply` function provided by the [`kfctl` package](https://github.com/kubeflow/kfctl/tree/master/pkg) to deploy Kubeflow. The Kubeflow resources specified with the manifests will be added with the following annotation to indicate that they are owned by this _KfDef_ instance.
+* When a _KfDef_ instance is created or updated, the operator's _reconciler_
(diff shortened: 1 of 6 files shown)


VaishnaviHire:
Adding `ownerReferrences` without changing the scope of the `KfDef` crd to `Cluster` is not possible from v1.20

Cross-namespace ownerreferences are [disallowed](https://github.com/kubernetes-sigs/controller-runtime/pull/675#) and dependent objects are subject to deletion. This PR needs to be revisited, once the scope of `KfDef` is updated.

